Issue #8, opened 2015-01-12
can not configure pool's org.killbill.dao.minIdle connections


kares:
sample KB configuration : 
```
JAVA_OPTS="${JAVA_OPTS} -Dorg.killbill.dao.maxActive=80"
JAVA_OPTS="${JAVA_OPTS} -Dorg.killbill.dao.minIdle=30"
```

KB's relevant logs and exception on startup : 

```
2015-01-12 11:30:13,274 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning default value [true] for [org.killbill.dao.cachePrepStmts] on [org.killbill.commons.jdbi.guice.DaoConfig#isPrep$
2015-01-12 11:30:13,275 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning default value [true] for [org.killbill.dao.useServerPrepStmts] on [org.killbill.commons.jdbi.guice.DaoConfig#is$
2015-01-12 11:30:13,276 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning null default value for [org.killbill.dao.dataSourceClassName] on [org.killbill.commons.jdbi.guice.DaoConfig#get$
2015-01-12 11:30:13,277 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning null default value for [org.killbill.dao.driverClassName] on [org.killbill.commons.jdbi.guice.DaoConfig#getDriv$
2015-01-12 11:30:13,277 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning default value [4.0] for [org.killbill.dao.mysqlServerVersion] on [org.killbill.commons.jdbi.guice.DaoConfig#get$
2015-01-12 11:30:13,277 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning default value [HIKARICP] for [org.killbill.dao.poolingType] on [org.killbill.commons.jdbi.guice.DaoConfig#getCo$
2015-01-12 11:30:13,278 [localhost-startStop-1] INFO  o.s.c.ConfigurationObjectFactory - Assigning default value [0m] for [org.killbill.dao.maxConnectionAge] on [org.killbill.commons.jdbi.guice.DaoConfig#getMax$
[...]

PR #11, opened 2015-02-24
support for dao.initSql + working dao.minIdle with HikariCP

kares:
mostly need a fix for #8 and the ability to set connection init SQL (mostly for the osgi pool)
... also removed BoneCP (not really production ready) - best to avoid

p.s. C3P0 might need to be updated to 0.9.5 for the init SQL to work

**TODO** update master pom dependencies - C3P0 (0.9.5 final), HikariCP and remove BoneCP

### DIFF
--- a/jdbi/pom.xml
+++ b/jdbi/pom.xml
@@ -41,11 +41,6 @@
             <artifactId>h2</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.jolbox</groupId>
-            <artifactId>bonecp</artifactId>
-            <optional>true</optional>
-        </dependency>
-        <dependency>
             <groupId>com.mchange</groupId>
             <artifactId>c3p0</artifactId>
             <optional>true</optional>
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
@@ -44,7 +44,7 @@ public interface DaoConfig {
 
     @Description("The minimum allowed number of idle connections to the database")
     @Config("org.killbill.dao.minIdle")
-    @Default("1")
+    @Default("0")
     int getMinIdle();
 
     @Description("The maximum allowed number of active connections to the database")
@@ -122,5 +122,5 @@ public interface DaoConfig {
     @Config("org.killbill.dao.poolingType")
     @Default("HIKARICP")
     DataSourceConnectionPoolingType getConnectionPoolingType();
-    
+
 }
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
@@ -73,6 +73,11 @@ public interface DaoConfig {
     @Default("5m")
     TimeSpan getIdleConnectionTestPeriod();
 
+    @Description("Sets a SQL statement executed after every new connection creation before adding it to the pool")
+    @Config("org.killbill.dao.connectionInitSql")
+    @DefaultNull
+    String getConnectionInitSql();
+
     @Description("Number of prepared statements that the driver will cache per connection")
     @Config("org.killbill.dao.prepStmtCacheSize")
     @Default("500")
@@ -117,4 +122,5 @@ public interface DaoConfig {
     @Config("org.killbill.dao.poolingType")
     @Default("HIKARICP")
     DataSourceConnectionPoolingType getConnectionPoolingType();
+    
 }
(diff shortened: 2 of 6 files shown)


pierre:
Interesting, from the BoneCP GitHub [README.md](https://github.com/wwadge/bonecp):

```
BoneCP is a Java JDBC connection pool implementation that is tuned for high performance
by minimizing lock contention to give greater throughput for your applications.
It beats older connection pools such as C3P0 and DBCP but should now be considered
deprecated in favour of HikariCP
```

We know at least one company running `BoneCP` in production, but given the mention above, it's probably fine to drop support moving forward.

kares:
@pierre in that case I can update the commits to keep BoneCP and print a deprecation that it will go.
pretty much assumed it's not used that much since I recall you were experimenting with (C3P0) pooling alternatives not that long ago ... your call here

pierre:
@kares I think it's fine, don't bother. This will land sometime in 0.13.x anyways, which is the development release train. We''ll add a note to the upgrade notes for 0.12.x to 0.14.x that BoneCP isn't supported anymore (https://github.com/killbill/killbill/wiki/Kill-Bill-version-migration-path).

FYI, we went back-and-forth between C3P0 and BoneCP over the years, and we always had issues with both. Things seem better with HikariCP now (**knock** on wood...).

kares:
some more updates here (making sure tests will run smooth with an updated HikariCP - 2.3.x does fail-fast connection validation by default) - also changed the *dao.minIdle* to default to **0** for pool impl consistency

relevant pom.xml changes are at https://github.com/kares/killbill-oss-parent/tree/pooling I understand this will probably have to wait or should I fire a PR there ? 
... for the record an updated HikariCP has better performance characteristic mostly observable by no longer starting an initialization thread every few seconds.

pierre:
> some more updates here (making sure tests will run smooth with an updated HikariCP - 2.3.x does fail-fast connection validation by default) - also changed the dao.minIdle to default to 0 for pool impl consistency

Thanks! I'll merge it on my branch and re-run the tests. It will be in 0.13.6.

> relevant pom.xml changes are at https://github.com/kares/killbill-oss-parent/tree/pooling I understand this will probably have to wait or should I fire a PR there ?

No worries - I've cherry-picked your changes here: https://github.com/killbill/killbill-oss-parent/tree/kares-performance.

> ... for the record an updated HikariCP has better performance characteristic mostly observable by no longer starting an initialization thread every few seconds.

I got side-tracked by a few bugs (https://github.com/killbill/killbill/issues/292, https://github.com/killbill/killbill/issues/293) that look jDBI related, but will compare the results with the new version.
I also need to figure out why a `purchase` call takes about 2s vs 500ms for `add_payment_method`. It looks like `Response#create_response_and_transaction ` (https://github.com/killbill/killbill-plugin-framework-ruby/blob/kares-performance/lib/killbill/helpers/active_merchant/active_record/models/response.rb#L51) takes most of the time - to be confirmed...

kares:
> I also need to figure out why a purchase call takes about 2s vs 500ms for add_payment_method. It looks like Response#create_response_and_transaction (https://github.com/killbill/killbill-plugin-framework-ruby/blob/kares-performance/lib/killbill/helpers/active_merchant/active_record/models/response.rb#L51) takes most of the time - to be confirmed...

main reason I introduced *dao.initSql* is for the OSGi pool to be set into an AR compatible way (on each connection checkout AR does `configure_connection` which does `SET` variables SQL) ... thus [avoid the query cost](https://github.com/jruby/activerecord-jdbc-adapter/blob/bc57bb1144b07721e99b5e6d045d2046ff135ffe/lib/arjdbc/mysql/adapter.rb#L44-L78) on every AR connection checkout (`with_connection_and_transaction`). 
although I do not have any numbers/testing backing this up I think it should help ... 

should also note that it seems that with HikariCP 2.3.3 support for initSQL got deprecated ;(

pierre:
> main reason I introduced dao.initSql is for the OSGi pool to be set into an AR compatible way (on each connection checkout AR does configure_connection which does SET variables SQL) ... thus avoid the query cost on every AR connection checkout (with_connection_and_transaction). 

I haven't tested the property yet but heads up that it looks like it doesn't support running multiple queries (https://github.com/brettwooldridge/HikariCP/issues/152). 

> should also note that it seems that with HikariCP 2.3.3 support for initSQL got deprecated ;(


It looks like they will keep it, see https://groups.google.com/forum/#!searchin/hikari-cp/connectionInitSql/hikari-cp/9n-N5W0FIOI/mpoH8Nbk9ZoJ:

> Now, circling back around to connectionInitSql.  We're going to retract @Deprecated from connectionInitSql for two primary reasons.  First, following the "simple things should be simple" maxim, the burden of implementing a custom class when all that is needed is a single statement execution is too high.  Second, from what we can see every major pool supports a similar construct, it is pure configuration, and therefore movement between pool implementations is not impeded.

kares:
> I haven't tested the property yet but heads up that it looks like it doesn't support running multiple queries

should still be fine for now ... variables can be wrapped (and actually are with AR) in one `SET` statement

PR #16, opened 2015-10-29
@TimedResource metrics

andrenpaes:
Intercept resources with `@TimedResource` annotation and time execution.
Creates a timer metric for annotated resource method with the following dimensions:
* Resource Path (from @Path annotation on the class)
* Name (either from the @TimedResource annotation or from the method name)
* Tags (from parameters annotated with @MetricTag)
* Http Method
* Response status group (2xx, 3xx, etc)

The unit tests cover the Guice Listener for creating interceptors and the Interceptor itself.

Here's an example of a generated metric name: `kb_resource./1.0/kb/accounts.getAccountByKey.GET.4xx.404`

Right now, this is different from the plugin generated metrics, since it doesn't have the `killbill-service` prefix, so this is something we could add to keep things consistent:
`killbill-service.kb_plugin_errors.killbill-adyen.PaymentPluginApi.getPaymentInfo`

Eventually, we should try consolidating the metric generation/naming in the `metrics` module, so it will be easier to manage this.

This PR is part of https://github.com/killbill/killbill/issues/385

### DIFF
--- a/skeleton/pom.xml
+++ b/skeleton/pom.xml
@@ -146,6 +146,10 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.kill-bill.commons</groupId>
+            <artifactId>killbill-metrics</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.skife.config</groupId>
             <artifactId>config-magic</artifactId>
             <optional>true</optional>
--- a/skeleton/pom.xml
+++ b/skeleton/pom.xml
@@ -90,7 +90,6 @@
         <dependency>
             <groupId>com.sun.jersey</groupId>
             <artifactId>jersey-server</artifactId>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>com.sun.jersey.contribs</groupId>
--- a/skeleton/pom.xml
+++ b/skeleton/pom.xml
@@ -90,6 +90,7 @@
         <dependency>
             <groupId>com.sun.jersey</groupId>
             <artifactId>jersey-server</artifactId>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>com.sun.jersey.contribs</groupId>
--- a/skeleton/src/main/java/org/killbill/commons/skeleton/metrics/ResourceTimer.java
+++ b/skeleton/src/main/java/org/killbill/commons/skeleton/metrics/ResourceTimer.java
@@ -16,11 +16,13 @@
 
 package org.killbill.commons.skeleton.metrics;
 
+import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import com.google.common.base.Joiner;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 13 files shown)


skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
```diff
@@ -29,33 +43,133 @@
...
+        }
+    }
+
+    private static String capitalize(String property) {
+        return property.substring(0, 1).toUpperCase() + property.substring(1);
+    }
+
+    private static IllegalArgumentException handleReadPropertyError(Object object, String property, Exception e) {
+        return new IllegalArgumentException(String.format("Failed to read tag property \"%s\" value from object of type %s", property, object.getClass()), e);
     }
```
andrenpaes: The above `getProperty` section is a bit ugly (thank you Java), but I didn't find anything in Guava that could help us there, and I also didn't want to add another dependency to commons-beanutils.

If you guys have any suggestion here, I'd love to get rid of this code.

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
```diff
@@ -29,33 +43,133 @@
...
-            if (result instanceof Response) {
-                status = ((Response) result).getStatus();
+            final Object response = invocation.proceed();
+            if (response instanceof Response) {
+                responseStatus = ((Response) response).getStatus();
+            } else if (response == null || response instanceof Void) {
+                responseStatus = Response.Status.NO_CONTENT.getStatusCode();
+            } else {
+                responseStatus = Response.Status.OK.getStatusCode();
             }
```
sbrossie: Is that true (if response is not a `Response` and != null and ! response instanceof Void => Status.OK) ?

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
```diff
@@ -29,33 +43,133 @@
...
-            return result;
-        } catch (final WebApplicationException ex) {
-            status = ex.getResponse().getStatus();
-            throw ex;
+
+            return response;
+        } catch (WebApplicationException e) {
+            responseStatus = e.getResponse().getStatus();
+
```
sbrossie: Are we sure `e.getResponse()` will not be null?

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
```diff
@@ -29,33 +43,133 @@
  */
 public class TimedResourceInterceptor implements MethodInterceptor {
 
-    private final TimedResourceMetric timer;
+    private final Provider<GuiceContainer> jerseyContainer;
+    private final Provider<MetricRegistry> metricRegistry;
+    private final String resourcePath;
```
sbrossie: Any reason why we don't keep track of `MetricRegistry` and `GuiceContainer ` instead of `Provider<GuiceContainer>` and `Provider<MetricRegistry>` ?

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
andrenpaes: Yeah, a resource can return something like a String, or any object really. In that case Jersey returns a 200 response and the object as the body.

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
andrenpaes: We are. `WebApplicationException` always have a response object. Even if it's created with a null response, it initializes one:
http://grepcode.com/file/repo1.maven.org/maven2/com.sun.jersey/jersey-core/1.18.3/javax/ws/rs/WebApplicationException.java#83

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptor.java
andrenpaes: The interceptor is created by the listener. The listener itself is create during Guice "configuration time" (??), when the actual instances of those classes are not yet available.
See this line: https://github.com/killbill/killbill-commons/pull/16/files#diff-9da97a8842ac4154b2c4c4ee8a5cb529R84

I've based this impl in the existing one, which was using Guice `SettableFuture`. I changed it to using `Provider`, since its api is a little bit easier to use.

I've tried using direct injection, but it would just give me `null`. Maybe there's some way around it?

skeleton/pom.xml
```diff
@@ -90,7 +90,6 @@
         <dependency>
             <groupId>com.sun.jersey</groupId>
             <artifactId>jersey-server</artifactId>
-            <scope>test</scope>
```
pierre: Why this change?

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/ResourceTimer.java
```diff
@@ -0,0 +1,77 @@
...
+                values.add(value);
+            } else {
+                values.add("null");
+            }
+        }
+        return values;
+    }
+
+    private String responseStatusGroup(int responseStatus) {
+        return String.format("%sxx", responseStatus / 100);
```
pierre: Is this to avoid generating too many metrics? I'm just wondering if it makes sense: for example, requests resulting in 201 are expected to take more time than 200.

skeleton/src/main/java/org/killbill/commons/skeleton/modules/StatsModule.java
```diff
@@ -80,9 +81,9 @@ protected void configure() {
         install(new AdminServletModule(healthCheckUri, metricsUri, pingUri, threadsUri));
 
         // Metrics/Jersey integration
-        final TimedResourceListener listener = new TimedResourceListener();
-        requestInjection(listener);
-        bindListener(Matchers.any(), listener);
+        final TimedResourceListener timedResourceTypeListener =
+                new TimedResourceListener(getProvider(GuiceContainer.class), getProvider(MetricRegistry.class));
```
pierre: Nit: are you going on a new line to limit the code width? I think it makes diffs hard to read.

skeleton/src/test/java/org/killbill/commons/skeleton/metrics/ResourceTimerTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+import com.google.common.collect.ImmutableMap;
+import org.testng.Assert;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+@Test(groups = "fast")
+public class ResourceTimerTest {
```
pierre: We prefix test classes with `Test`.

skeleton/pom.xml
andrenpaes: This package contains the GuiceContainer class, which is used for the exception mapping.
See: https://github.com/killbill/killbill-commons/pull/16/files#diff-a2c6a89915677e225f2f5b804057314bR46

skeleton/src/test/java/org/killbill/commons/skeleton/metrics/TimedResourceInterceptorTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+import com.codahale.metrics.MetricRegistry;
+import com.google.inject.AbstractModule;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.google.inject.matcher.Matchers;
+import com.sun.jersey.guice.spi.container.servlet.GuiceContainer;
+
+@Test(groups = "fast")
+public class TimedResourceInterceptorTest {
```
pierre: Same here (`TestTimedResourceInterceptor` instead).

skeleton/src/main/java/org/killbill/commons/skeleton/metrics/ResourceTimer.java
pierre: Oh, scratch that, I see we would get `2xx.201`.

PR #19, opened 2017-01-18
Support unpooled database connections

rhanneken:
Currently, Kill Bill supports the use of either [c3p0](http://www.mchange.com/projects/c3p0/) or [HikariCP](https://brettwooldridge.github.io/HikariCP/) for pooling database connections. In either case, database connections are managed by software running in the same JVM that Kill Bill is
running in.

It would be nice to have the option of configuring Kill Bill so that it doesn't try to pool database connections. This would make it easier to use external database connection pools like [PGBouncer](http://pgbouncer.github.io/).

See also [killbill-platform pull request 18](https://github.com/killbill/killbill-platform/pull/18).

### DIFF
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
@@ -18,6 +18,7 @@
 
 package org.killbill.commons.jdbi.guice;
 
+import java.io.IOException;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
@@ -25,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 
 import javax.sql.DataSource;
 
+import org.killbill.commons.embeddeddb.EmbeddedDB;
 import org.skife.config.TimeSpan;
 
 import com.codahale.metrics.MetricRegistry;
@@ -48,6 +50,8 @@ public class DataSourceProvider implements Provider<DataSource> {
 
     private Object metricRegistry;
     private Object healthCheckRegistry;
+
+    private EmbeddedDB embeddedDB;
 
     @VisibleForTesting
     static enum DatabaseType {
@@ -90,22 +94,36 @@ public class DataSourceProvider implements Provider<DataSource> {
         this.healthCheckRegistry = healthCheckRegistry;
     }
 
+    @Inject(optional = false)
+    public void setEmbeddedDB(final EmbeddedDB embeddedDB) {
+        this.embeddedDB = embeddedDB;
+    }
+
     @Override
     public DataSource get() {
         return buildDataSource();
     }
 
     private DataSource buildDataSource() {
-        if (DataSourceConnectionPoolingType.C3P0.equals(config.getConnectionPoolingType())) {
-            loadDriver();
-            return new C3P0DataSourceBuilder().buildDataSource();
-        }
-
-        if (DataSourceConnectionPoolingType.HIKARICP.equals(config.getConnectionPoolingType())) {
-            if (dataSourceClassName != null) {
+        switch(config.getConnectionPoolingType()) {
+            case C3P0:
                 loadDriver();
-            }
-            return new HikariDataSourceBuilder().buildDataSource();
+                return new C3P0DataSourceBuilder().buildDataSource();
+            case HIKARICP:
+                if (dataSourceClassName != null) {
+                    loadDriver();
+                }
+                return new HikariDataSourceBuilder().buildDataSource();
+            case NONE:
+                try {
+                    embeddedDB.initialize();
+                    embeddedDB.start();
+                    return embeddedDB.getDataSource();
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            default:
+                break;
         }
 
         throw new IllegalArgumentException("DataSource " + config.getConnectionPoolingType() + " unsupported");
--- a/jdbi/pom.xml
+++ b/jdbi/pom.xml
@@ -148,6 +148,10 @@
         </dependency>
         <dependency>
             <groupId>org.kill-bill.commons</groupId>
+            <artifactId>killbill-embeddeddb-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.kill-bill.commons</groupId>
             <artifactId>killbill-embeddeddb-h2</artifactId>
             <scope>test</scope>
         </dependency>
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceConnectionPoolingType.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceConnectionPoolingType.java
@@ -20,5 +20,6 @@ package org.killbill.commons.jdbi.guice;
 
 public enum DataSourceConnectionPoolingType {
     C3P0,
-    HIKARICP
+    HIKARICP,
+    NONE
 }
--- a/jdbi/src/test/java/org/killbill/commons/jdbi/guice/TestDataSourceProvider.java
+++ b/jdbi/src/test/java/org/killbill/commons/jdbi/guice/TestDataSourceProvider.java
@@ -43,6 +43,10 @@ import java.util.logging.Logger;
 
 import javax.sql.DataSource;
 
+import org.h2.jdbcx.JdbcConnectionPool;
+import org.killbill.commons.embeddeddb.EmbeddedDB;
+import org.killbill.commons.embeddeddb.GenericStandaloneDB;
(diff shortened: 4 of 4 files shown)


sbrossie:
@rhanneken The patch looks good. Did you have a chance to verify that all three pools ('shiro', 'main', 'osgi') end up with the right the 'no pooling' Datasource?

jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
```diff
@@ -230,6 +243,57 @@ DataSource buildDataSource() {
...
+            }
+
+            if (DatabaseType.MYSQL.equals(databaseType)) {
+                Reference ref = new Reference(dataSourceClassName);
+                ref.add(new StringRefAddr("user", config.getUsername()));
+                ref.add(new StringRefAddr("password", config.getPassword()));
+                ref.add(new StringRefAddr("explicitUrl", "true"));
+                ref.add(new StringRefAddr("url", config.getJdbcUrl()));
+                try {
+                    return (DataSource) new MysqlDataSourceFactory().getObjectInstance(ref, null, null, null);
```
pierre: Have you looked at re-using parts of HikariCP instead? Something like a custom `UnpooledDataSource` which calls `ProxyFactory.getProxyConnection` (see `BaseHikariPool`) to create connections?

I agree that it's a bit odd to use part of the pooling code to implement an unpooled DataSource but among advantages:

* it would prevent us from having to hardcode all the per-driver properties (e.g. it seems that `serverName` is missing for MySQL)
* no need to pull all drivers in killbill-jdbi
* no need for MariaDB vs MySQL custom configuration (we would return `com.mysql.jdbc.jdbc2.optional.MysqlDataSource` here but we actually use `org.mariadb.jdbc.MySQLDataSource` by default today)

rhanneken: Re using HikariCP: I'm not sure how that would work.  [`ProxyFactory.getProxyConnection`](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-2.3.7/hikaricp-common/src/main/java/com/zaxxer/hikari/proxy/ProxyFactory.java#L50) requires that one pass a [`HikariPool`](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-2.3.7/hikaricp/src/main/java/com/zaxxer/hikari/pool/HikariPool.java) object as an argument.  `HikariPool` is a `final` class whose methods I can't override to behave the way I want them to.

Or are you saying I should use the Javassist library to [specify my own implementations for the `ProxyFactory` methods](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-2.3.7/hikaricp/src/main/java/com/zaxxer/hikari/proxy/JavassistProxyFactory.java#L95-L120) at runtime?  I'm not clear on how that works, but I could look into it if you think that's a promising approach.

Re MySQL's `serverName` property:  since I provide the `url` property, am I not allowed to omit `serverName`?  I admit I haven't tested that yet.

pierre: > `HikariPool` is a `final` class whose methods I can't override to behave the way I want them to.

Ah indeed, I overlooked this 😿 

> Or are you saying I should use the Javassist library to [specify my own implementations for the `ProxyFactory` methods](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-2.3.7/hikaricp/src/main/java/com/zaxxer/hikari/proxy/JavassistProxyFactory.java#L95-L120) at runtime?  I'm not clear on how that works, but I could look into it if you think that's a promising approach.

Nah, that's probably too much of a headache.

> Re MySQL's `serverName` property:  since I provide the `url` property, am I not allowed to omit `serverName`?  I admit I haven't tested that yet.

My remark was more about the risk of missing or misconfiguring the `DataSource`, rather than this specific property. We have several pools and layers of `DataSource` in the platform code, and I'm always nervous whenever we need to touch this, especially when we need to instantiate/manipulate drivers code directly. Hence my pushback to see if there is a way we could simply this.

Let me suggest one more route.

Re-reading the code, `DataSourceProvider` is injected via `ReferenceableDataSourceSpyProvider` in [KillbillPlatformModule](https://github.com/killbill/killbill-platform/blob/master/server/src/main/java/org/killbill/billing/server/modules/KillbillPlatformModule.java#L115) and in [DefaultOSGIModule](https://github.com/killbill/killbill-platform/blob/master/osgi/src/main/java/org/killbill/billing/osgi/glue/DefaultOSGIModule.java#L86). We already have access to [EmbeddedDB](https://github.com/killbill/killbill-commons/blob/master/embeddeddb/common/src/main/java/org/killbill/commons/embeddeddb/EmbeddedDB.java) (confusing name - it's really just an abstraction for the DB) in the first module (where it's bound) and it's easy to pass it to `DefaultOSGIModule` (because this module is initiated from `KillbillPlatformModule`).

This means that `EmbeddedDB` could be injected to `DataSourceProvider` and could provide the `DataSource` for free, including both MySQL and MariaDB (`initialize` might need to be called manually though, TBD). In case of PostgreSQL, the code would return `PGSimpleDataSource` for instance. Would that work?

pierre: @rhanneken Did you have a chance to review my proposal? If not, no worries, I'm just trying to get a sense if we should wait on this for 0.18.3.

sbrossie: @rhanneken Any update? We 'd like to release `0.18.3` (we have a few bug fixes for issues you guys reported) and we would like to know if we should include your change for DB pooling?

jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
```diff
@@ -90,22 +94,36 @@ public void setHealthCheckRegistry(final HealthCheckRegistry healthCheckRegistry
         this.healthCheckRegistry = healthCheckRegistry;
     }
 
+    @Inject(optional = false)
```
pierre: Shouldn't it be `true` (only required for `NONE` pooling type)? We could simply add a NPE check in the corresponding branch in `buildDataSource` below.

pierre:
> One thing makes me worry: I call `start` on EmbeddedDB, but I never call `stop`.  Does that violate the class' contract? 

It does, but there is no easy way to fix it at the moment unfortunately (we already have the issue when using [embedded H2 for instance](https://github.com/killbill/killbill-platform/blob/ca3f9161b70c90fc301ef24c51d4491d02b07b33/server/src/main/java/org/killbill/billing/server/modules/EmbeddedDBProvider.java#L58)). In practice though, this should only be used in conjunction with a `GenericStandaloneDB`, so there isn't any side effect of not calling `stop`.

Were you able to run tests with `PgBouncer`? If so, do you see any impact on performance?

rhanneken:
@sbrossie @pierre I discovered a problem with my previous code change; because it was using the EmbeddedDB singleton defined in KillbillPlatformModule, our plugin code was connecting with the wrong user/schema.  To support separate `DataSource`s for Kill Bill and its plugins, I had to make more changes.  I also had to make [changes to killbill-platform](https://github.com/killbill/killbill-platform/issues/18).

The changes feel a bit ugly to me, but early testing suggests everything is functionally correct.  We still need to do some proper performance testing.

pierre:
@rhanneken I've adapted your changes and pushed them into the `unpooled` branch of both repos. Could you give it a try to confirm it's working as expected?

rhanneken:
@pierre `DataSourceProvider` is saying "java.lang.IllegalArgumentException: DataSource NONE unsupported" when `buildDataSource` is called.  I think this is because `embeddedDB` is null, maybe because `KillbillPlatformModule.provideDataSourceInAComplicatedWayBecauseOf627` is called before `KillbillPlatformModule.configureEmbeddedDB`.

Even if that worked, it seems to me that you have one embeddedDB configured with the main `DaoConfig`.  I don't think that would work for plugins, which need the data source configured with the `OSGIDataSourceConfig` configuration.

pierre:
> `DataSourceProvider` is saying "java.lang.IllegalArgumentException: DataSource NONE unsupported" when `buildDataSource` is called.

Sorry, I forgot to push the Kill Bill changes. See 09eb9f2255f390363125c0b6e95d5c7c0b6a3e55.

> Even if that worked, it seems to me that you have one embeddedDB configured with the main `DaoConfig`.  I don't think that would work for plugins, which need the data source configured with the `OSGIDataSourceConfig` configuration.

Good catch! Fixed by 64c23a73780475e5f397dd90da32b57f360419a8.

PR #23, opened 2017-09-27
Import queue retries code from Kill Bill

pierre:
To simplify the diff, `work-for-release-0.19.x` should be brought up-to-date first.

See https://github.com/killbill/killbill/issues/767.


queue/src/main/java/org/killbill/queue/retry/RetryableSubscriber.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public interface SubscriberAction<T extends BusEvent> {
+
+        void run(T event);
+    }
+
+    public static final class SubscriberQueueHandler implements NotificationQueueHandler {
+
+        // Similar to com.google.common.eventbus.SubscriberRegistry
+        private static final LoadingCache<Class<?>, ImmutableSet<Class<?>>> FLATTEN_HIERARCHY_CACHE =
```
sbrossie: If we were to only implement retry around `ExtBusEvent` (plugin only), then we would have only one type `org.killbill.billing.beatrix.extbus.DefaultBusExternalEvent`, and probably we could simplify quite a bit here

pierre: Yes, probably, but we still want the retry piece for invoice plugins, no?

queue/src/main/java/org/killbill/queue/retry/RetryableService.java
```diff
@@ -0,0 +1,157 @@
...
+        try {
+            final NotificationEvent retryNotificationEvent = new RetryNotificationEvent(objectMapper.writeValueAsString(originalNotificationEvent), originalNotificationEvent.getClass(), originalEffectiveDate, retryNb);
+            retryNotificationQueue.recordFutureNotification(effectiveDate, retryNotificationEvent, userToken, searchKey1, searchKey2);
+        } catch (final IOException e) {
+            log.error("Unable to schedule retry for event='{}', effectiveDate='{}'", originalNotificationEvent, effectiveDate, e);
+        }
+    }
+
+    private DateTime computeRetryDate(final QueueRetryException invoicePluginApiRetryException, final DateTime initialEventDateTime, final int retryNb) {
+        final List<Period> retrySchedule = invoicePluginApiRetryException.getRetrySchedule();
```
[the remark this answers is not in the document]
pierre: Ah, yes, leftover from the prototype.

Fixed by 0e438713c9e3293f4dd1f61f282f57cf8a9d7929.

PR #26, opened 2017-11-14
jdbi: add new config property org.killbill.dao.initializationFailFast

pierre:
By default, HikariCP shuts down the pool if it encounters an error during initialization.
Guice swallows this error, Kill Bill stops its startup sequence and no information is returned to the user.

This patch introduces a new config property to control this behavior and changes the default:

* If false (new default), the startup sequence continues. Subsequent attempts to retrieve a connection through `HikariDataSource.getConnection` will log an exception if any (the Kill Bill startup sequence continues nonetheless)
* If true (previous behavior), the connection exception is now logged but Guice swallows it and Kill Bill stops its startup sequence

This new default should help debug connectivity issues when provisioning cloud environments.

@sbrossie Could you test this patch to see if it either works around your timing issues or if it helps debug it?

### DIFF
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
@@ -132,4 +132,9 @@ public interface DaoConfig {
     @Config("org.killbill.dao.healthCheckExpected99thPercentile")
     @Default("50ms")
     TimeSpan getHealthCheckExpected99thPercentile();
+
+    @Description("Whether or not initialization should fail on error immediately")
+    @Config("org.killbill.dao.initializationFailFast")
+    @Default("false")
+    boolean isInitializationFailFast();
 }
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
@@ -28,6 +28,8 @@ import javax.sql.DataSource;
 
 import org.killbill.commons.embeddeddb.EmbeddedDB;
 import org.skife.config.TimeSpan;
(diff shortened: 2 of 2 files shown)


jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
```diff
@@ -132,4 +132,9 @@
     @Config("org.killbill.dao.healthCheckExpected99thPercentile")
     @Default("50ms")
     TimeSpan getHealthCheckExpected99thPercentile();
+
+    @Description("Whether or not initialization should fail on error immediately")
+    @Config("org.killbill.dao.initializationFailFast")
+    @Default("false")
+    boolean isInitializationFailFast();
```
sbrossie: Don't we want to default to true ?

jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
```diff
@@ -200,7 +206,13 @@ DataSource buildDataSource() {
                 }
             }
 
-            return new HikariDataSource(hikariConfig);
+            try {
+                return new HikariDataSource(hikariConfig);
+            } catch (final PoolInitializationException e) {
+                // When initializationFailFast=true, log the exception to alert the user (the Guice initialization sequence will continue though)
```
sbrossie: > Guice initialization sequence will continue though

Maybe something we should address-- i think leaving the stack up in that state, is not so great.

jdbi/src/main/java/org/killbill/commons/jdbi/guice/DaoConfig.java
pierre: `true` is the current behavior -- it is less cloud friendly (timing issues).

PR #27, opened 2017-11-22
ZooKeeper Implementation

javier-gomez-github:
- Added Dependency for ZooKeeper
- Created new ZooKeeperGlobalLock and Locker
- Imported WriteLock recipes from ZooKeeper and modified

### DIFF
--- a/locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLock.java
+++ b/locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLock.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2017 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.killbill.commons.locker.zookeeper;
+
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.ZooDefs;
+import org.apache.zookeeper.ZooKeeper;
+import org.apache.zookeeper.data.ACL;
+import org.killbill.commons.locker.GlobalLock;
+import org.killbill.commons.locker.zookeeper.recipes.LockListener;
+import org.killbill.commons.locker.zookeeper.recipes.WriteLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ZooKeeperGlobalLock implements GlobalLock {
+    private static final Logger LOG = LoggerFactory.getLogger(ZooKeeperGlobalLock.class);
+
+    private final String name;
+    private final String path;
+    private final WriteLock writeLock;
+    private final CountDownLatch lockAcquiredSignal = new CountDownLatch(1);
+
+    private static final List<ACL> DEFAULT_ACL = ZooDefs.Ids.OPEN_ACL_UNSAFE;
+
+    public ZooKeeperGlobalLock(String name, ZooKeeper zookeeper, String path) {
+        this(name, zookeeper, path, DEFAULT_ACL);
+    }
+
+    public ZooKeeperGlobalLock(String name, ZooKeeper zookeeper, String path, List<ACL> acl) {
+        this.name = name;
+        this.path = path;
+        writeLock = new WriteLock(zookeeper, path, acl, new SyncLockListener());
+    }
+
+    public void lock() throws InterruptedException, KeeperException {
+        LOG.debug("{} requesting lock on {}...", name, path);
+        writeLock.lock();
+        lockAcquiredSignal.await();
+    }
+
+    public boolean lock(long timeout, TimeUnit unit) throws InterruptedException, KeeperException {
+        LOG.debug("{} requesting lock on {} with timeout {} {}...", name, path, timeout, unit);
+        writeLock.lock();
+        return lockAcquiredSignal.await(timeout, unit);
+    }
+
+    public boolean tryLock() throws InterruptedException, KeeperException {
+        return lock(1, TimeUnit.SECONDS);
+    }
+
+    public void unlock() {
+        writeLock.unlock();
+    }
+
+    @Override
+    public void release() {
+        // TODO jgomez: define actions for this method (is it the same as the lockReleased() method below?)
+    }
+
+    class SyncLockListener implements LockListener {
+        @Override
+        public void lockAcquired() {
+            LOG.debug("Lock acquired by {} on {}", name, path);
+            lockAcquiredSignal.countDown();
+        }
+
+        @Override
+        public void lockReleased() {
+            LOG.debug("Lock released by {} on {}", name, path);
+        }
+    }
+}
--- a/locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLock.java
+++ b/locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLock.java
@@ -16,75 +16,16 @@
 
 package org.killbill.commons.locker.zookeeper;
 
-import java.util.List;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
+import java.sql.Connection;
 
-import org.apache.zookeeper.KeeperException;
-import org.apache.zookeeper.ZooDefs;
-import org.apache.zookeeper.ZooKeeper;
-import org.apache.zookeeper.data.ACL;
 import org.killbill.commons.locker.GlobalLock;
-import org.killbill.commons.locker.zookeeper.recipes.LockListener;
-import org.killbill.commons.locker.zookeeper.recipes.WriteLock;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.killbill.commons.locker.GlobalLockBase;
+import org.killbill.commons.locker.GlobalLockDao;
+import org.killbill.commons.locker.ResetReentrantLockCallback;
 
-public class ZooKeeperGlobalLock implements GlobalLock {
-    private static final Logger LOG = LoggerFactory.getLogger(ZooKeeperGlobalLock.class);
+public class ZooKeeperGlobalLock extends GlobalLockBase implements GlobalLock {
 
-    private final String name;
-    private final String path;
-    private final WriteLock writeLock;
-    private final CountDownLatch lockAcquiredSignal = new CountDownLatch(1);
-
-    private static final List<ACL> DEFAULT_ACL = ZooDefs.Ids.OPEN_ACL_UNSAFE;
-
-    public ZooKeeperGlobalLock(String name, ZooKeeper zookeeper, String path) {
-        this(name, zookeeper, path, DEFAULT_ACL);
-    }
-
-    public ZooKeeperGlobalLock(String name, ZooKeeper zookeeper, String path, List<ACL> acl) {
-        this.name = name;
-        this.path = path;
-        writeLock = new WriteLock(zookeeper, path, acl, new SyncLockListener());
-    }
-
-    public void lock() throws InterruptedException, KeeperException {
-        LOG.debug("{} requesting lock on {}...", name, path);
-        writeLock.lock();
-        lockAcquiredSignal.await();
-    }
-
-    public boolean lock(long timeout, TimeUnit unit) throws InterruptedException, KeeperException {
-        LOG.debug("{} requesting lock on {} with timeout {} {}...", name, path, timeout, unit);
-        writeLock.lock();
-        return lockAcquiredSignal.await(timeout, unit);
-    }
-
-    public boolean tryLock() throws InterruptedException, KeeperException {
-        return lock(1, TimeUnit.SECONDS);
-    }
-
-    public void unlock() {
-        writeLock.unlock();
-    }
-
-    @Override
-    public void release() {
-        // TODO jgomez: define actions for this method (is it the same as the lockReleased() method below?)
-    }
-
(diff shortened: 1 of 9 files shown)


locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLocker.java
```diff
@@ -0,0 +1,66 @@
...
+                @Override
+                public void process(final WatchedEvent event) {
+                    if (event.getState() == Watcher.Event.KeeperState.SyncConnected) {
+                        connectedSignal.countDown();
+                    }
+                }
+            });
+            // TODO jgomez 2: what should be the Lock Name?
+            // TODO jgomez 3: where do I get the accountId from commons?
+            return new ZooKeeperGlobalLock("ZooKeeper Lock", zooKeeper, zkPath + "<account_id>");
```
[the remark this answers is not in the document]
javier-gomez-github: Cool!
And what about the ZooKeeper Path? 
I'm trying to build it as `zkPath` + `accountId` (as Stephane suggested). Being `zkPath = "/kb/locks/accounts/"`
But the accountId is part of the `lockName`, right? How is the `lockName` being built so I can extract the `accountId`

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLocker.java
```diff
@@ -0,0 +1,66 @@
...
+import org.killbill.commons.locker.GlobalLockerBase;
+import org.killbill.commons.locker.ResetReentrantLockCallback;
+
+public class ZooKeeperGlobalLocker extends GlobalLockerBase implements GlobalLocker {
+
+    final CountDownLatch connectedSignal = new CountDownLatch(1);
+    protected static final String zkPath = "/kb/locks/accounts/";
+
+    public ZooKeeperGlobalLocker() {
+        super(null, null, DEFAULT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
javier-gomez-github: Agree with the naming! 😕

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLock.java
```diff
@@ -0,0 +1,90 @@
...
+        this(name, zookeeper, path, DEFAULT_ACL);
+    }
+
+    public ZooKeeperGlobalLock(String name, ZooKeeper zookeeper, String path, List<ACL> acl) {
+        this.name = name;
+        this.path = path;
+        writeLock = new WriteLock(zookeeper, path, acl, new SyncLockListener());
+    }
+
+    public void lock() throws InterruptedException, KeeperException {
```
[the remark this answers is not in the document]
javier-gomez-github: so basically the ZooKeeperGlobalLock and ZooKeeperGlobalLocker will be very similar to the ones for MySql and PostgreSql right?
All the logic will happen in the ZooKeeperGlobalLockDao as mentioned below?

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLocker.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    protected GlobalLock getGlobalLock(final Connection connection, final String lockName, final ResetReentrantLockCallback resetCb) {
+        try {
+            // TODO jgomez 1: use the same timeout value?
+            ZooKeeper zooKeeper = new ZooKeeper("connectString", Long.valueOf(timeout).intValue(), new Watcher() {
```
sbrossie: Couple of remarks here:
1. I (would) think initialization of `ZooKeeper` should not be done each time we get the lock, but rather in the constructor. Now, that requires checking that 1/ such objects can safely be shared among different threads and 2/ Make sure there is code to handle cases when connection breaks, to re-initialize the object. In a first prototype implementation this could be omitted for simplicity.

2. The use the the `CountDownLatch` `connectedSignal` does not seem correct. The whole intent is to synchronize the main thread in this method with the one handling the callback. Therefore this main thread should call `await()`:

```
public ZooKeeper connect(String hosts, int sessionTimeout)
        throws IOException, InterruptedException {
  final CountDownLatch connectedSignal = new CountDownLatch(1);
  ZooKeeper zk = new ZooKeeper(hosts, sessionTimeout, new Watcher() {
    @Override
    public void process(WatchedEvent event) {
      if (event.getState() == Watcher.Event.KeeperState.SyncConnected) {
        connectedSignal.countDown();
      }
    }
  });
  connectedSignal.await();
  return zk;
}

```

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLocker.java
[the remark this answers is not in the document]
sbrossie: I suggested `zkPath + accountId` but what i meant was really  `zkPath + lockName` -- the `lockName` will internally be built on the `accountId`.

locker/src/main/java/org/killbill/commons/locker/zookeeper/recipes/ProtocolSupport.java
```diff
@@ -0,0 +1,186 @@
+/*
```
sbrossie: For all these recipes files:
1. In there a small jar we could import instead of copying the classes -- if there is none, maybe this is ok.
2. If we go with copying classes, we should **keep the original copyright** . Also we should document the version so we can later update such files if needed.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
```diff
@@ -0,0 +1,107 @@
...
+import org.killbill.commons.locker.zookeeper.recipes.WriteLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ZooKeeperGlobalLockDao implements GlobalLockDao {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ZooKeeperGlobalLockDao.class);
+    private static final List<ACL> DEFAULT_ACL = ZooDefs.Ids.OPEN_ACL_UNSAFE;
+
+    // TODO jgomez: we can read this property from the external properties
```
sbrossie: I think this is probably fine to keep it hardcoded.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
```diff
@@ -0,0 +1,107 @@
...
+    private final String path;
+    private final WriteLock writeLock;
+    private final CountDownLatch lockAcquiredSignal = new CountDownLatch(1);
+
+    public ZooKeeperGlobalLockDao(final String lockName, final Long timeout) {
+        this.name = lockName;
+        this.path = zkPath + lockName;
+        ZooKeeper zooKeeper = null;
+        try {
+            // TODO jgomez: what should we use as the connectString value? Using localhost by now
```
sbrossie: Once you implement the config, you 'll need additional properties to specify this. Just keep the `TODO` for now.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
```diff
@@ -0,0 +1,107 @@
...
+            zooKeeper = new ZooKeeper("localhost", timeout.intValue(), new Watcher() {
+                @Override
+                public void process(final WatchedEvent event) {
+                    if (event.getState() == Event.KeeperState.SyncConnected) {
+                        lockAcquiredSignal.countDown();
+                    }
+                }
+            });
+            lockAcquiredSignal.await();
```
sbrossie: 1. I think it would make sense to have timeout in the wait() method -- probably the same one or slightly higher than the one above.

2. Also i would isolate this initialization code in a routine that you call here, in such a way that if/when connection breaks, you can reinitialize by calling this same routine. 

3. Code to detect when connection breaks, so connection can be re-initialized can come later, but definitely something we will need to handle.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
```diff
@@ -0,0 +1,107 @@
...
+        } catch (Exception e) {
+            // TODO jgomez catch and throw corresponding exceptions
+        }
+        this.writeLock = new WriteLock(zooKeeper, path, DEFAULT_ACL, new SyncLockListener());
```
sbrossie: This `writeLock` object cannot be shared, and needs to be allocated for each lock we are trying to take. In particular it uses a `SyncLockListener` which itself contains a `CountDownLatch` to notify the thread waiting in the `lock()` method.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    class SyncLockListener implements LockListener {
+        @Override
+        public void lockAcquired() {
+            LOG.debug("Lock acquired by {} on {}", name, path);
+            lockAcquiredSignal.countDown();
```
sbrossie: We need a different `CountDownLatch` here, that is used on a per lock level and shared between the `SyncLockListener` and the `lock()` method. I would think, creating all these objects inside the `lock()` method below should work. So something like the following -- code may need to be tweaked a bit:

```
public boolean lock(final Connection connection, final String lockName, final long timeout, final TimeUnit timeUnit) throws SQLException {
        
    private final CountDownLatch signal = new CountDownLatch(1);
    final WriteLock writeLock = new WriteLock(zooKeeper, path, DEFAULT_ACL, new SyncLockListener(signal));
    
    try {
        LOG.debug("{} requesting lock on {}...", lockName, path);
        writeLock.lock();
        return signal.await(timeout, timeUnit);
    } catch (Exception e) {
        // TODO jgomez catch and throw corresponding exceptions
        return false;
    }
}
 ```

File: locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java [hunk 11]
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public boolean releaseLock(final Connection connection, final String lockName) throws SQLException {
+        // TODO jgomez: implement! Is this the unlock method?
+        writeLock.unlock();
+        return false;
+    }
+
+    @Override
+    public boolean isLockFree(final Connection connection, final String lockName) throws SQLException {
+        // TODO jgomez: implement! How to check this in the writeLock?
```
sbrossie: Either the `WriteLock`class already offers such method -- did not verify -- or we will need to implement one. Most of the logic should already be in place though in this class

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
javier-gomez-github: @sbrossie could you explain a little bit more this situation? 
Remember that my distributed synchronisation skills are not the best 😐

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
javier-gomez-github: And what about the unlock() method? Which writeLock should I use?
Should I maintain a global writeLock and use it in the unlock() method?
Sorry I'm a little confused here.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
javier-gomez-github: WriteLock has a ´isOwner()´ method that says if the current Node is the owner of the Lock.
Is that what we are expecting here?

sbrossie:
@javier-gomez-github Most of your questions above are tied to understanding the basic mechanism, so let me try to explain:

The distributed lock aspect is implemented in the [WriteLock](https://github.com/apache/zookeeper/blob/master/src/recipes/lock/src/java/org/apache/zookeeper/recipes/lock/WriteLock.java) on top of  Zookeeper. For each lock operation, you should allocate a new object `WriteLock`. In the [CTOR](https://github.com/apache/zookeeper/blob/master/src/recipes/lock/src/java/org/apache/zookeeper/recipes/lock/WriteLock.java#L78) you will pass the following:
* An instance of Zookeeper -- some global object that is shared and represents the connection to zookeeper cluster
* A callback `LockListener` that is also assigned for each lock operation. The callback is an interface that defines `lockAcquired()` and which is called when the lock got acquired. 

So, in the end your problem is not one of distributed locking but more simply of synchronizing two threads. Look at the [blog post](https://nofluffjuststuff.com/blog/scott_leberknight/2013/07/distributed_coordination_with_zookeeper_part_5_building_a_distributed_lock), once you **understand the model**, pretty much all you have to do is copy the code and adjust it. Based in this blog post you can see that:

1. There is a `CountDownLatch` latch called  `signal` that is allocated on a **per request basis** and that is shared between the main thread calling the `lock()` method and the thread that calls the `LockListener#lockAcquired` method when the operation has completed.
2. When the  `LockListener#lockAcquired` is called,  this results in calling `signal.countDown()` and so your main thread in the lock now returns from the call `signal.await()`, which means that it now holds the 'distributed lock' and can proceed with whatever it needs to do.
3. Finally, this main thread will call `unlock()`; the [WriteLock implementation](https://github.com/apache/zookeeper/blob/master/src/recipes/lock/src/java/org/apache/zookeeper/recipes/lock/WriteLock.java#L137) also allows to ensure that your main thread only returns from such operation when this is complete using the same  callback mechanism but this is not really needed -- the log post does not use that, and i don't think we need it either.

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
sbrossie: @javier-gomez-github  : Don't implement it for now. We can come back to it later

locker/src/main/java/org/killbill/commons/locker/zookeeper/ZooKeeperGlobalLockDao.java
sbrossie: We don't need unlock -- this is my point `3.` below. Focus on the `lock()` implementation for now.

PR #28, opened 2017-12-13
Support for queues healthcheck

pierre:


skeleton/src/main/java/org/killbill/commons/skeleton/modules/StatsModule.java
```diff
@@ -86,19 +86,8 @@ protected void configure() {
...
-    protected HealthCheckRegistry provideHealthCheckRegistry(final Set<HealthCheck> healthChecks) {
-        final HealthCheckRegistry healthCheckRegistry = createHealthCheckRegistry();
-
-        // It used to be done by AdminServletProvider. JavaDoc of AdminServletModule outdated?
-        for (final HealthCheck healthCheck : healthChecks) {
-            healthCheckRegistry.register(healthCheck.getClass().getName(), healthCheck);
-        }
 
-        return healthCheckRegistry;
+        bind(HealthCheckRegistry.class).toInstance(createHealthCheckRegistry());
```
sbrossie: So if i understand correctly, we bind this `HealthCheckRegistry` to a method that return a `new KillBillHealthCheckRegistry ()` (default CTOR), at which point such object does not do anything. Later Guice finds out we annotated the method `initialize` and does the method injection, but at that time DBI was initialized ands things work.

queue/src/main/java/org/killbill/queue/DBBackedQueue.java
```diff
@@ -521,13 +521,15 @@ public Long apply(final T input) {
...
     private long getNbReadyEntries() {
         final Date now = clock.getUTCNow().toDate();
-        final String owner = CreatorName.get();
-        return sqlDao.getNbReadyEntries(now, owner, config.getTableName());
+        return getNbReadyEntries(now);
     }
 
+    public long getNbReadyEntries(final Date now) {
+        final String owner = config.getPersistentQueueMode() == PersistentQueueMode.POLLING ? null : CreatorName.get();
+        return sqlDao.getNbReadyEntries(now, owner, config.getTableName());
```
sbrossie: In the case of `STICKY_EVENTS` we should be able to get info from the `inflightEvents` queue instead of doing the query?

skeleton/src/main/java/org/killbill/commons/skeleton/modules/StatsModule.java
pierre: Yup! The registry needs to be created and passed to the DAO (HikariCP healthcheck) first. Then, when the DAO is setup, the healthchecks themselves can be initialized and injected into the registry.

queue/src/main/java/org/killbill/queue/DBBackedQueue.java
pierre: Yes, but only if the queue didn't overflow. It's a rare case but since we only perform the query once every 5 minutes, not sure the optimization is worth it?

PR #35, opened 2018-02-21
jdbi: optimize LowerToCamelBeanMapper

pierre:
Cache transformation detection at initialization time, to avoid costly isAssignableFrom() calls at runtime.


pierre:
Yup, pretty much.

PR #36, opened 2018-02-23
jdbi: perf updates

pierre:
* Make sure DBI caches are used for the queues APIs
* Ensure prepared statements are cached (server-side) by default (regression with recent MariaDB driver upgrade)

### DIFF
--- a/embeddeddb/mysql/src/main/java/org/killbill/commons/embeddeddb/mysql/KillBillMariaDbDataSource.java
+++ b/embeddeddb/mysql/src/main/java/org/killbill/commons/embeddeddb/mysql/KillBillMariaDbDataSource.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


embeddeddb/mysql/src/main/java/org/killbill/commons/embeddeddb/mysql/KillBillMariaDbDataSource.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        if (url != null) {
+            setUrl(buildUpdatedUrl(url));
+        }
+    }
+
+    // No easy way to set MariaDB options at runtime besides updating the JDBC url
+    @VisibleForTesting
+    String buildUpdatedUrl(final String url) throws SQLException {
```
sbrossie: Interesting. I am wondering if we should not even go one step further, as suggested by the HikariCP  link above:

```
dataSource.useLocalSessionState=true
dataSource.useLocalTransactionState=true
dataSource.rewriteBatchedStatements=true
dataSource.cacheResultSetMetadata=true
dataSource.cacheServerConfiguration=true
dataSource.elideSetAutoCommits=true
dataSource.maintainTimeStats=false
```

Also this post may be interested for this:  https://www.facebook.com/notes/mysql-at-facebook/jdbc-connection-overheads/10152027251270933/

pierre: It looks like most of these only apply to the MySQL driver, not to the [MariaDB one](https://mariadb.com/kb/en/library/about-mariadb-connector-j/), except for `rewriteBatchedStatements` (which we don't want, as it overrides `useServerPrepStmts` to `false`).

That being said, there might be further work to be done to reduce roundtrips (`SET NAMES`, `SHOW VARIABLES`, etc.).

Issue #29, opened 2018-01-26
STICKY_POLLING optimizations for the cloud


pierre:
Current behavior for `STICKY_POLLING` configuration, for both the bus and notification queue, is to only look at entries in the database that match the `Creator` name of the current node. The issue is when a node goes away: currently, a manual update of the database entries is required (see https://github.com/killbill/killbill/wiki/Kill-Bill-Bus-and-Notification-Queue-Configuration).

Instead, we would now have a new `Reaper` thread that periodically monitors such entries and re-dispatch them automatically. To detect such entries, we need to look for bus events with an old created date or notification events with an effective date too far in the past (actual thresholds configurable, 5' by default). Such detection is already done today in the queries (see second part of the `where` clause in https://github.com/killbill/killbill-commons/blob/624a8190780f1587d9f78472c2be0b3a36ea27c4/queue/src/main/resources/org/killbill/bus/dao/PersistentBusSqlDao.sql.stg#L14 and https://github.com/killbill/killbill-commons/blob/624a8190780f1587d9f78472c2be0b3a36ea27c4/queue/src/main/resources/org/killbill/notificationq/dao/NotificationSqlDao.sql.stg#L21), we would need to introduce similar queries but focusing only on such _old_ entries.  These entries would then be updated to reflect the current `Creator` name (maybe re-dispatch 10 at a time, configurable).

Additional tests to write in `TestDBBackedQueue`.

PR #38, opened 2018-02-28
 STICKY_POLLING optimizations for the cloud #29

wwjbatista:


queue/src/main/resources/org/killbill/queue/dao/QueueSqlDao.sql.stg
```diff
@@ -207,6 +207,19 @@ updateOnError(tableName) ::= <<
...
+updateCreatingOwner(tableName) ::= <<
+    update <tableName>
+    set
+      creating_owner = :newCreatingOwner
+      , processing_owner = NULL
+      , processing_available_date = :now
+      , processing_state = 'AVAILABLE'
+    where
+      <reapWhereClause()>
+    limit :max
```
pierre: I believe this won't work on PostgreSQL unfortunately (could be the cause of the CI PostgreSQL failure).

queue/src/main/resources/org/killbill/queue/dao/QueueSqlDao.sql.stg
```diff
@@ -207,6 +207,19 @@ updateOnError(tableName) ::= <<
     ;
 >>
 
+updateCreatingOwner(tableName) ::= <<
+    update <tableName>
+    set
+      creating_owner = :newCreatingOwner
+      , processing_owner = NULL
+      , processing_available_date = :now
```
pierre: I believe `processing_available_date` should be `NULL` instead, as it is used as a threshold at which nodes can steal entries from other nodes in a non-STICKY mode (cc @sbrossie for confirmation).

queue/src/main/resources/org/killbill/bus/dao/PersistentBusSqlDao.sql.stg
```diff
@@ -18,6 +18,13 @@ readyOrderByClause() ::= <<
     record_id asc
 >>
 
+reapWhereClause() ::= <<
```
pierre: Isn't there a race condition where the same entries will be re-dispatched over and over again, until they are processed? In other words, if the polling of the bus is slower than the polling of the reaper, the same re-dispatched entries will be seen as old again, and re-dispatched.

The only option around it might be to update the `created_date` (cc @sbrossie).

queue/src/main/java/org/killbill/queue/DBBackedQueue.java
```diff
@@ -756,4 +756,14 @@ public Long execute() throws RuntimeException {
             }
         });
     }
+
+    public void updateCreatingOwner(final Date reapingDate) {
+        sqlDao.inTransaction(new Transaction<Void, QueueSqlDao<T>>() {
```
pierre: One more comment: it would be good to add some logging, probably `WARN` level, in case we do update some rows -- this should be uncommon.

queue/src/main/java/org/killbill/bus/api/PersistentBusConfig.java
```diff
@@ -92,4 +92,16 @@
     @Default("bus_events_history")
     @Description("Bus events history table name")
     public abstract String getHistoryTableName();
+
+    @Override
+    @Config("org.killbill.persistent.bus.${instanceName}.reapThreshold")
+    @Default("5m")
```
sbrossie: I would change the default to at least `10m` (i know i had suggested `5m` on the ticket); Re-reading the code, `5m` is also the default time entries are being claimed, so this `reapThreshold` default value should be greater than the default `claim.time`.

queue/src/main/resources/org/killbill/queue/dao/QueueSqlDao.sql.stg
sbrossie: @pierre @wwjbatista yes, `processing_available_date` should be set to null.

Also, couple of things:
1. We should order things to start from older entries (valid if the `max` is less than total number)
2. It would be good to keep track of which entries where moved and this goes back to @pierre 's comment about logging.  At the very least, the update will return the number of rows updated, we could log that.

@pierre @wwjbatista For for thoughts -- not yet actionable:
Should we extend our schema to add an `updated_date` and also a `prev_creating_owner`.  The `prev_creating_owner` would be mostly interesting to understand what happens. My worry is that we are trying to add complexity around an error case, and so when this happens, verifying things worked as expected might be good.

queue/src/main/resources/org/killbill/bus/dao/PersistentBusSqlDao.sql.stg
sbrossie: @pierre Agreed

Couple of options:
1. We configuring our reaping delay large enough but there is no guarantee : If there is a backup, potentially we end up in that state, and that will only make the problem worst
2. We change the design: Each reaper threads checks that 'things are not moving' for each other node before trying to re-balance entries. Since node don't know each other, that would require keeping a `last_record_id` processed for each existing nodes; i think this is feasible but adds some complexity ....

wwjbatista: How about a new `reaped_date` defaulted to null?.  Every time an entry is reaped the date is set. If the reaped date is not null then it could be reap.  That way we could conserve the created_date and know that it was reaped.

pierre: Yeah, I was thinking hard to find a solution without changing the schema: `bus_events_history` is usually massive (hundreds of millions of rows) in large deployments, so changing it is not always an option. But I don't think there is a reliable way to handle this race condition otherwise.

On the plus side, if we end up changing it, we could add more information as suggested by @sbrossie above.

That being said, I think looking at `reaped_date != NULL` by itself isn't enough. Consider the scenario where a node reaps an entry, updates `reaped_date` to `now()` and then crashes before processing that reaped entry... 

@wwjbatista Let me have a chat with @sbrossie before making any change on this specifically (you can still go ahead and address the other comments) and we'll update here with how complicated we want this piece of code to be...

queue/src/main/java/org/killbill/bus/api/PersistentBusConfig.java
pierre: 👍 

Should we also ensure in the code that `claim.time < reapThreshold` and override `reapThreshold` if its mis-configured (with a `WARN` log)?

queue/src/main/resources/org/killbill/bus/dao/PersistentBusSqlDao.sql.stg
pierre: @wwjbatista Here is what we are going to do... In a transaction:

* Move the current entry into the history table, but update the state to `REAPED` (`PersistentQueueEntryLifecycleState` to be updated)
* Insert a new entry into the main table, which is a copy of the previous entry, but with an updated `created_date`

queue/src/main/java/org/killbill/queue/DefaultReaper.java
```diff
@@ -0,0 +1,118 @@
...
+            reapEntriesHandle.cancel(true);
+        }
+
+        if (!scheduler.isShutdown()) {
+            scheduler.shutdown();
+            try {
+                scheduler.awaitTermination(5, TimeUnit.SECONDS);
+            } catch (final InterruptedException e) {
+                Thread.currentThread().interrupt();
+                log.info(String.format("%s stop sequence has been interrupted",threadScheduledExecutorName));
```
pierre: You need to log first, then interrupt the thread.

Also, with SLF4J, you actually don't need `String.format`, you can do:

```
log.info("{} stop sequence has been interrupted",threadScheduledExecutorName);
```

You need to use `{}` and not `%s` though.

queue/src/main/java/org/killbill/queue/DefaultReaper.java
```diff
@@ -0,0 +1,118 @@
...
+        return isStarted.get();
+    }
+
+    private long getReapThreshold() {
+        final long threshold;
+        // if Claim time is greater than reap threshold
+        if (config.getClaimedTime().getMillis() >= config.getReapThreshold().getMillis()) {
+            // override reap threshold using claim time + 5 minutes
+            threshold = config.getClaimedTime().getMillis() + FIVE_MINUTES;
+            log.warn(String.format("%s: Reap threshold was mis-configured. Claim time [%s] is greater than reap threshold [%s]",
```
pierre: Same remark regarding `String.format`.

queue/src/main/java/org/killbill/queue/DBBackedQueue.java
```diff
@@ -763,47 +764,33 @@ public void reapEntries(final Date reapingDate) {
...
-                        set(entry, "createdDate", clock.getUTCNow());
-                        set(entry, "processingState", PersistentQueueEntryLifecycleState.AVAILABLE);
+                        entry.setCreatedDate(now);
+                        entry.setProcessingState(PersistentQueueEntryLifecycleState.AVAILABLE);
+                        entry.setCreatingOwner(CreatorName.get());
                         insertEntryFromTransaction(transactional, entry);
                     }
 
-                    log.warn(String.format( "%s %s entries were reaped by %s",DB_QUEUE_LOG_ID ,entriesLeftBehind.size(), CreatorName.get()));
+                    log.warn(String.format( "{} {} entries were reaped by {}",DB_QUEUE_LOG_ID ,entriesLeftBehind.size(), CreatorName.get()));
```
pierre: @wwjbatista Here and below: you also need to remove the `String.format` call:

```
log.warn("{} {} entries were reaped by {}",DB_QUEUE_LOG_ID ,entriesLeftBehind.size(), CreatorName.get());
 ```

See https://www.slf4j.org/faq.html#logging_performance.

Issue #30, opened 2018-01-26
STICKY_EVENTS optimization for the cloud


pierre:
~To make https://github.com/killbill/killbill-commons/issues/29 work with the `STICKY_EVENTS` configuration (bus only), we also need to review the behavior of finding orphan entries.~

~Instead of simply logging:
https://github.com/killbill/killbill-commons/blob/624a8190780f1587d9f78472c2be0b3a36ea27c4/queue/src/main/java/org/killbill/queue/DBBackedQueue.java#L346~

~we now need to re-insert such orphan `ids` in the in-memory `inflightEvents` queue.~

Looking at the [code](https://github.com/killbill/killbill-commons/pull/40#discussion_r173397400) written for `STICKY_POLLING`, there is an easier approach to this:

* Keep the orphan code as-is
* In the `reapEntries` method, instead of simply calling `transactional.insertEntries`, do something similar to `insertEntryFromTransaction`, i.e. write the entries to disk and add them to the inflight queue in bulk on success. Because of the `LAST_INSERT_ID` logic, we wouldn't be able to insert them in bulk in that case though

For testing:
* Additional tests to write in `TestDBBackedQueue`
* Manual testing: you could have one Kill Bill server running in `STICKY_EVENTS` and insert a dummy bus event entry for another fictive node (other creator name). That event should eventually be picked up by the server

PR #40, opened 2018-03-08
 Issue#29 - notificationq reaper

wwjbatista:


queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
```diff
@@ -41,6 +41,7 @@
 import org.killbill.queue.DBBackedQueue;
 import org.killbill.queue.InTransaction;
 import org.killbill.queue.QueueObjectMapper;
+import org.killbill.queue.api.PersistentQueueConfig.PersistentQueueMode;
```
pierre: Extra import.

queue/src/main/java/org/killbill/queue/DBBackedQueue.java
```diff
@@ -783,10 +783,11 @@ public T apply(@Nullable final T entry) {
                         entry.setCreatedDate(now);
                         entry.setProcessingState(PersistentQueueEntryLifecycleState.AVAILABLE);
                         entry.setCreatingOwner(CreatorName.get());
-                        insertEntryFromTransaction(transactional, entry);
                     }
 
-                    log.warn(String.format( "{} {} entries were reaped by {}",DB_QUEUE_LOG_ID ,entriesLeftBehind.size(), CreatorName.get()));
+                    transactional.insertEntries(entriesLeftBehind, config.getTableName());
```
pierre: @wwjbatista I think for this PR (`STICKY_POLLING`), bypassing the logic from `insertEntryFromTransaction` going straight to the DB is fine, although we should at least update the stats (`totalInsert.inc()`).

@sbrossie But I don't think this is what we want for the full implementation (i.e. considering `STICKY_EVENTS` https://github.com/killbill/killbill-commons/issues/30) as we wouldn't re-insert the `ids` into the inflight queue (unless we want to solely rely on the orphan polling mechanism?).

PR #41, opened 2018-03-12
Add debug context

pierre:
* clock: log time when we reset the clock
* queue: log full event in DEBUG entries
* queue: put the user token in the MDC

### DIFF
--- a/queue/src/main/java/org/killbill/queue/dispatching/Dispatcher.java
+++ b/queue/src/main/java/org/killbill/queue/dispatching/Dispatcher.java
@@ -1,6 +1,6 @@
 /*
- * Copyright 2015 Groupon, Inc
- * Copyright 2015 The Billing Project, LLC
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
  *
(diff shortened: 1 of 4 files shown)


queue/src/main/java/org/killbill/queue/dispatching/Dispatcher.java
```diff
@@ -94,25 +100,35 @@ public CallableQueue(final M entry, final CallableCallback<E, M> callback) {
...
-                long errorCount = entry.getErrorCount();
-                try {
-                    callback.dispatch(event, entry);
-                } catch (final Exception e) {
-                    if (e.getCause() != null && e.getCause() instanceof InvocationTargetException) {
-                        lastException = e.getCause().getCause();
-                    } else {
+            try {
+                final UUID userToken = entry.getUserToken();
+                MDC.put(MDC_KB_USER_TOKEN, userToken != null ? userToken.toString() : null);
```
sbrossie: So, will this be logged automatically or does this require changes in the log format as well?

pierre: To close the loop on this: https://github.com/killbill/killbill/pull/896#discussion_r174279072.

PR #45, opened 2018-03-19
queue: add NotificationQueue#removeFutureNotificationsForSearchKeys API

pierre:
This is for https://github.com/killbill/killbill/issues/864.

### DIFF
--- a/queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
+++ b/queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
@@ -342,9 +342,11 @@ public class DefaultNotificationQueue implements NotificationQueue {
         dao.getSqlDao().inTransaction(new Transaction<Void, QueueSqlDao<NotificationEventModelDao>>() {
             @Override
             public Void inTransaction(final QueueSqlDao<NotificationEventModelDao> transactional, final TransactionStatus status) throws Exception {
+                // Move entries by batch into the history table
                 final int batchSize = 25;
                 final Collection<NotificationEventModelDao> currentBatch = new ArrayList<NotificationEventModelDao>(batchSize);
 
+                // Note that we don't claim them here, so it could be possible that some of these entries end up being processed nonetheless
                 final Iterator<NotificationEventModelDao> futureQueueEntriesForSearchKeys = ((NotificationSqlDao) transactional).getReadyQueueEntriesForSearchKeys(getFullQName(),
                                                                                                                                                                    searchKey1,
                                                                                                                                                                    searchKey2,
--- a/queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
+++ b/queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
@@ -20,6 +20,8 @@ package org.killbill.notificationq;
 
 import java.io.IOException;
 import java.sql.Connection;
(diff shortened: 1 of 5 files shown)


queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
```diff
@@ -327,6 +337,44 @@ public Void withSqlDao(final NotificationSqlDao transactional) throws Exception
...
+    @Override
+    public void removeFutureNotificationsForSearchKeys(final Long searchKey1, final Long searchKey2) {
+        dao.getSqlDao().inTransaction(new Transaction<Void, QueueSqlDao<NotificationEventModelDao>>() {
+            @Override
+            public Void inTransaction(final QueueSqlDao<NotificationEventModelDao> transactional, final TransactionStatus status) throws Exception {
+                // Move entries by batch into the history table
+                final int batchSize = 25;
+                final Collection<NotificationEventModelDao> currentBatch = new ArrayList<NotificationEventModelDao>(batchSize);
+
+                // Note that we don't claim them here, so it could be possible that some of these entries end up being processed nonetheless
```
sbrossie: But then, wouldn't we get an exception when the thread processing the entry is trying to move the row into `history`. Is this api only used for tests ?

queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
```diff
@@ -327,6 +337,44 @@ public Void withSqlDao(final NotificationSqlDao transactional) throws Exception
         InTransaction.execute(dbi, connection, handler, NotificationSqlDao.class);
     }
 
+    @Override
+    public void removeFutureNotificationsForSearchKeys(final Long searchKey1, final Long searchKey2) {
+        dao.getSqlDao().inTransaction(new Transaction<Void, QueueSqlDao<NotificationEventModelDao>>() {
+            @Override
+            public Void inTransaction(final QueueSqlDao<NotificationEventModelDao> transactional, final TransactionStatus status) throws Exception {
+                // Move entries by batch into the history table
+                final int batchSize = 25;
```
sbrossie: Should we return an indicator to the user saying that there are more entries than the default batch size, or if not, how will she know that she should call back the api?

pierre: All entries would be removed, this is just to call `dao.moveEntriesToHistoryFromTransaction` by batch (`@SqlBatch`), instead of one at a time. See https://github.com/killbill/killbill-commons/pull/45/files/c2b601f7093ff881a9cd3b7efe029f4820208335#diff-0b9d631efc836141ed593b2514afb5e8R361.

queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
pierre: > But then, wouldn't we get an exception when the thread processing the entry is trying to move the row into `history`.

I don't think so, as `moveEntriesToHistory` is really:

1. Add the entries from memory (already selected) into the history table
2. Delete by `record_id` in the main table

So in the worst case scenario, we could have duplicates in the history table (2 rows with 2 different states, `PROCESSED` and `REMOVED`), but I don't think we care?

> Is this api only used for tests ?

No, this would be called by the `closeAccount` endpoint.

PR #48, opened 2018-05-03
 Add the possibility to bypass mapping registry cache

wwjbatista:
killbill/killbill#335
killbill/killbill#972

### DIFF
--- a/jdbi/src/main/java/org/skife/jdbi/v2/MappingRegistry.java
+++ b/jdbi/src/main/java/org/skife/jdbi/v2/MappingRegistry.java
@@ -56,8 +56,7 @@ class MappingRegistry
 
     public ResultSetMapper mapperFor(Class type, StatementContext ctx) {
         // check if cache must be bypassed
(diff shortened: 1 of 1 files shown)


jdbi/src/main/java/org/skife/jdbi/v2/MappingRegistry.java
```diff
@@ -54,6 +55,10 @@ public void add(ResultSetMapperFactory factory)
     }
 
     public ResultSetMapper mapperFor(Class type, StatementContext ctx) {
+        // check if cache must be bypassed
+        Map<String, Object> attributes = ctx.getAttributes();
+        Boolean bypassCache = attributes.containsKey("bypassMappingRegistryCache") ? (Boolean) ctx.getAttribute("bypassMappingRegistryCache") : Boolean.FALSE;
```
pierre: Maybe `Boolean.valueOf` to avoid `ClassCastException`.

Issue #46, opened 2018-03-26
Missing userToken from RetryableService json


sbrossie:
The `originalEvent` from RetryableService is missing the `userToken`. Example:

```
{
  "originalEvent": "{\"busEvent\":{\"extBusEvent\":{\"objectId\":null,\"objectType\":\"INVOICE\",\"eventType\":\"INVOICE_NOTIFICATION\",\"accountId\":\"ba135655-1de0-4847-abcf-a4f0343fc733\",\"tenantId\":\"d05188d5-a9ee-4f07-908c-6174c437973a\",\"metaData\":null},\"extBusEventClass\":\"org.killbill.billing.beatrix.extbus.DefaultBusExternalEvent\",\"userToken\":null,\"searchKey1\":4184783402330769479,\"searchKey2\":5787557447408176903},\"busEventClass\":\"org.killbill.billing.osgi.KillbillEventRetriableBusHandler$OSGIBusEvent\"}",
  "originalEventClass": "org.killbill.queue.retry.SubscriberNotificationEvent",
  "originalEffectiveDate": "2018-03-26T04:53:18.000Z",
  "retryNb": 5
```

pierre:
@sbrossie I think we already have that information.

Here's an example of a failed notification with two retries:

```
*************************** 1. row ***************************
                record_id: 1
               class_name: org.killbill.notificationq.TestNotificationQueue$TestNotificationKey
               event_json: {"value":"Foo"}
               user_token: 916a3558-c84d-46d5-a23e-c6c0ba67dd66
             created_date: 2018-05-14 14:35:10
           creating_owner: acme
         processing_owner: acme
processing_available_date: 2018-05-14 14:35:14
         processing_state: FAILED
              error_count: 0
              search_key1: 65
              search_key2: 34
               queue_name: svc:queueName
           effective_date: 2018-05-14 14:35:12
        future_user_token: dc06ebbd-eb32-4161-8568-38d137c8aa8b
*************************** 2. row ***************************
                record_id: 2
               class_name: org.killbill.queue.retry.RetryNotificationEvent
               event_json: {"originalEvent":"{\"value\":\"Foo\"}","originalEventClass":"org.killbill.notificationq.TestNotificationQueue$TestNotificationKey","originalEffectiveDate":"2018-05-14T14:35:12.000Z","retryNb":1}
               user_token: dc06ebbd-eb32-4161-8568-38d137c8aa8b
             created_date: 2018-05-14 14:35:14
[...]

PR #49, opened 2018-05-11
queue: introduce specific states for user initiated retries

pierre:
The original event to be retried will be mark as RETRIED.

Subsequent retry events (org.killbill.queue.retry.RetryNotificationEvent) will also
be marked as RETRIED, except the last one (GIVEN_UP).

Error count will always be 0 for all of these.

### DIFF
--- a/queue/src/test/java/org/killbill/bus/TestRetries.java
+++ b/queue/src/test/java/org/killbill/bus/TestRetries.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2014-2018 Groupon, Inc
+ * Copyright 2014-2018 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


pierre:
@sbrossie As discussed, I've implemented the changes:

* Removed the new states:
 * Original queue event is marked as `FAILED` if retried
 * Retries are marked as `FAILED` or `PROCESSED` (last one)
* Added a test for bus events

Note that the original queue event isn't moved to `PROCESSED` if the final retry worked, as this would require an API change to `NotificationQueueHandler`. I'm not sure this is what we want anyways, as the information can be obtained by following tokens (https://github.com/killbill/killbill-commons/issues/46#issuecomment-388846904).

queue/src/test/java/org/killbill/bus/TestRetries.java
```diff
@@ -0,0 +1,193 @@
...
+                    return Iterators.size(busSqlDao.getHistoricalQueueEntriesForSearchKeys(SEARCH_KEY_1, SEARCH_KEY_2, persistentBusConfig.getHistoryTableName())) == 1 &&
+                           Iterators.size(notificationSqlDao.getHistoricalQueueEntriesForSearchKeys("notifications-retries:myEvent-listener", SEARCH_KEY_1, SEARCH_KEY_2, notificationQueueConfig.getHistoryTableName())) == 4;
+                }
+            });
+
+            // Initial event was processed once
+            historicalEntriesForOriginalEvent = ImmutableList.<BusEventModelDao>copyOf(busSqlDao.getHistoricalQueueEntriesForSearchKeys(SEARCH_KEY_1, SEARCH_KEY_2, persistentBusConfig.getHistoryTableName()));
+            Assert.assertEquals(historicalEntriesForOriginalEvent.size(), 1);
+            Assert.assertEquals((long) historicalEntriesForOriginalEvent.get(0).getErrorCount(), (long) 0);
+            // State is still FAILED
```
sbrossie: Ideally it should be `PROCESSED` , but i remember this is what you explained and good enough probably

PR #51, opened 2018-07-10
On tests; start fails after a stop is issued

wwjbatista:


wwjbatista:
@pierre; @sbrossie;  I found out the error running the test against `email-notification-plugin` with the `travis` profile. The problem was the  `stop` method and the field alias. With those change the test pass without problem. I changed the test `TestEmailNotificationServlet` to be MySql specific to comply with `killbill-oss-parentversion 0.141.97`.

pierre:
> I changed the test `TestEmailNotificationServlet` to be MySql specific to comply with `killbill-oss-parent version 0.141.97`. 

@wwjbatista What do you mean by `comply`? I'm not quite sure what you are referring to.

wwjbatista:
@pierre, to not fail the test,  since the version of `killbill-commons` that resides in `killbill-oss-parent version 0.141.97` does not have the fix.

pierre:
Oh, of course 🤦‍♂️ 😆 

Sounds good!

PR #58, opened 2018-10-24
queue: create jdbi-shaded artifact

pierre:
Create an artifact with our forked version of JDBI shaded. That way, projects can depend on it without having to worry about a conflicting JDBI version.

Note: make sure to exclude the killbill-jdbi transitive dependency.

### DIFF
--- a/queue/pom.xml
+++ b/queue/pom.xml
@@ -174,4 +174,36 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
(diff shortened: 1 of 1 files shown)


queue/pom.xml
```diff
@@ -174,4 +174,36 @@
...
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
+                            <shadedClassifierName>jdbi-shaded</shadedClassifierName>
+                            <relocations>
+                                <relocation>
+                                    <pattern>org.skife.jdbi</pattern>
+                                    <shadedPattern>org.killbill.queue.org.skife.jdbi</shadedPattern>
```
sbrossie: That does not require changing any kb code (e.g `import`) to use the new path?

pierre: Nope, it creates a new shaded artifact -- it doesn't replace the main one.

PR #60, opened 2019-02-12
jdbi: batch execute and generate keys

pierre:
This is an adaptation of the upstream code.


jdbi/src/main/java/org/skife/jdbi/v2/sqlobject/BatchHandler.java
```diff
@@ -1,6 +1,4 @@
 /*
- * Copyright (C) 2004 - 2014 Brian McCallister
- *
```
sbrossie: lol...

PR #61, opened 2019-02-15
Refactor DBBackedQ + update metrics + fix bugs

sbrossie:


queue/src/test/java/org/killbill/notificationq/TestNotificationQueue.java
```diff
@@ -382,7 +392,7 @@ public boolean apply(final Boolean input) {
      *
      * @throws Exception
      */
-    @Test(groups = "slow")
+    @Test(groups = "slow", enabled=false)
```
pierre: Are these old tests (why disabled)? Should we keep these around?

queue/src/main/java/org/killbill/notificationq/NotificationQueueDispatcher.java
```diff
@@ -164,17 +166,18 @@ public boolean isStarted() {
...
+            return new DispatchResultMetrics(0, -1);
         }
         log.debug("Notifications from {} to process: {}", config.getTableName(), notifications);
 
         for (final NotificationEventModelDao cur : notifications) {
             dispatcher.dispatch(cur);
         }
-        return notifications.size();
+        // No need to return time, this is easy to compute from caller
+        return new DispatchResultMetrics(notifications.size(), -1);
```
pierre: Returning `-1` is a bit confusing though because below we call `dispatchTime.update(metricsResult.getTimeNanoSec(), TimeUnit.NANOSECONDS);` regardless of the value. It works because notification isn't sticky but it's a bit sneaky...

queue/src/main/java/org/killbill/queue/DBBackedQueueWithInflightQueue.java
```diff
@@ -0,0 +1,327 @@
...
+                Thread.currentThread().interrupt();
+                log.warn("{} Got interrupted", DB_QUEUE_LOG_ID);
+                return 0;
+            }
+        }
+        return pollSleepTime;
+    }
+
+    @Override
+    public ReadyEntriesWithMetrics<T> getReadyEntries() {
```
pierre: `ReadyEntriesWithMetrics` is only used for Bus with inflight queue. Is it worth the complexity computing it here? Couldn't the caller always compute the time spent (like in other cases)?

queue/src/main/java/org/killbill/queue/DBBackedQueueWithInflightQueue.java
```diff
@@ -0,0 +1,327 @@
...
+
+        try {
+            // Add entry in the inflightQ and clear threadlocal
+            final Iterator<Long> entries = transientInflightQRowIdCache.iterator();
+            while (entries.hasNext()) {
+                final Long entry = entries.next();
+                final boolean result = inflightEvents.offer(entry);
+                if (result) {
+                    log.debug("{} Inserting entry {} into inflightQ", DB_QUEUE_LOG_ID, entry);
+                    // Q overflowed, which means we will stop writing entries into the Q, and as a result, we will end up stop reading
```
pierre: Is this comment still valid?

queue/src/main/java/org/killbill/queue/DBBackedQueueWithPolling.java
```diff
@@ -0,0 +1,220 @@
...
+            @Override
+            public Integer execute(final QueueSqlDao<T> queueSqlDao) {
+                long ini = System.nanoTime();
+                final Integer result = queueSqlDao.claimEntries(recordIds, clock.getUTCNow().toDate(), CreatorName.get(), nextAvailable, config.getTableName());
+                rawClaimEntriesTime.update(System.nanoTime() - ini, TimeUnit.NANOSECONDS);
+                return result;
+            }
+        });
+        // We should ALWAYS see the same number since we are in STICKY_POLLING mode and there is only one thread claiming entries.
+        // We keep the 2 cases below for safety (code was written when this was MT-threaded), and we log with warn (will eventually remove it in the future)
```
pierre: We should maybe start cleaning all of that stuff, it's a lot of codepaths 😅

queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
```diff
@@ -152,11 +197,17 @@ private void sleepALittle(final long loopTimeMsec) throws InterruptedException {
...
+        executor.shutdown();
         try {
             executor.awaitTermination(config.getPollingSleepTimeMs(), TimeUnit.SECONDS);
         } catch (final InterruptedException e) {
             log.info(String.format("%s: Stop sequence has been interrupted", svcQName));
+        } finally {
+            int remainingCompleted = completedOrFailedEvents.size();
+            int remainingRetried = retriedEvents.size();
+            if (remainingCompleted > 0 || remainingRetried > 0) {
+                log.warn(String.format("%s: Stopped queue with %d event/notifications non completed ", svcQName, (remainingCompleted + remainingRetried)));
```
pierre: Nit: no need for the `String.format`.

queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
```diff
@@ -82,66 +103,90 @@ public void run() {
...
+            private void sleepSporadically(final long loopTimeMsec) throws InterruptedException {
+                if (isStickyEvent) {
+                    // In this mode, the main thread does not sleep, but blocks on the inflightQ to minimize latency.
                     return;
                 }
-                final long remainingSleepTime = config.getPollingSleepTimeMs() - loopTimeMsec;
-                if (remainingSleepTime > 0) {
-                    Thread.sleep(remainingSleepTime);
+
+                final long MAX_SLEEP_TIME_MS = 100;
```
pierre: Nit: could be `static`.

queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
```diff
@@ -82,66 +103,90 @@ public void run() {
...
                 }
+                return curSize;
             }
 
-            private void sleepALittle(final long loopTimeMsec) throws InterruptedException {
-                if (config.getPersistentQueueMode() == PersistentQueueConfig.PersistentQueueMode.STICKY_EVENTS) {
-                    // Disregard config.getPollingSleepTimeMs() in that mode in case this is not correctly configured with 0
+            private void sleepSporadically(final long loopTimeMsec) throws InterruptedException {
+                if (isStickyEvent) {
+                    // In this mode, the main thread does not sleep, but blocks on the inflightQ to minimize latency.
```
pierre: Thought: could we change the code to be always in sticky events?

1. Get rid of the polling mode for bus
2. For notifications, have a separate thread that fetches the ready events and put them in the "inflight" queue, for the main dispatcher thread to pull from

I don't think it would change the overall behavior, but it could simplify the code a bit further?

queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
```diff
@@ -78,56 +81,64 @@
...
     @Inject
     public DefaultPersistentBus(@Named(QUEUE_NAME) final IDBI dbi, final Clock clock, final PersistentBusConfig config, final MetricRegistry metricRegistry, final DatabaseTransactionNotificationApi databaseTransactionNotificationApi) {
-        super("Bus", config);
+        super("Bus", config, metricRegistry);
         this.dbi = (DBI) dbi;
         this.clock = clock;
         this.config = config;
         this.dbBackedQId = "bus-" + config.getTableName();
-        this.dao = new DBBackedQueue<BusEventModelDao>(clock, dbi, PersistentBusSqlDao.class, config, dbBackedQId, metricRegistry, databaseTransactionNotificationApi);
+        this.dao = config.getPersistentQueueMode() == PersistentQueueMode.STICKY_EVENTS ?
```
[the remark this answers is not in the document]
sbrossie: The infligtQ is robust, but tied to nodes not crashing down... or tied to our reaper logic -- for which we have verified that basic scenario works. 

Now, the `POLLING` -- or at the very least `STICKY_POLLING` will be needed for notifications, and the logic is shared, so it probably would not simplify too much?

queue/src/test/java/org/killbill/notificationq/TestNotificationQueue.java
sbrossie: Absolutely, mistake, will fix.

queue/src/main/java/org/killbill/notificationq/NotificationQueueDispatcher.java
sbrossie: Yes i agree, but i would then need to do another change of interface to pass the metrics data from`NotificationQueueDispatcher#getReadyNotifications()`, and since i could easily avoid propagating this data because it's easy to compute in the client side, i made this 'hack'. I am also not so happy about it - but thought it was not worth creating more java classes, ... to pass this data.

queue/src/main/java/org/killbill/queue/DBBackedQueueWithInflightQueue.java
sbrossie: No, impossible to compute from client because the blocking happens against the inflightQ deep in this class. This is different from notificationQ though -- previous remark about the `-1`.

queue/src/main/java/org/killbill/queue/DBBackedQueueWithInflightQueue.java
sbrossie: Nope, will remove...

queue/src/main/java/org/killbill/queue/DBBackedQueueWithPolling.java
sbrossie: Ok, let's have the discussion - not 100% convinced, but i also like the idea of more simplicity.

queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
sbrossie: I started to make the change, but i realized it's all over this code - is this worth it?

queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
sbrossie: Yes will do.

queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
sbrossie: Let's talk..

queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
```diff
@@ -138,39 +149,44 @@ public void start() {
...
             dispatcher.start();
             startQueue();
-        }
 
-        if (config.getPersistentQueueMode() == PersistentQueueMode.STICKY_POLLING || config.getPersistentQueueMode() == PersistentQueueMode.STICKY_EVENTS) {
-            reaper.start();
+            if (config.getPersistentQueueMode() == PersistentQueueMode.STICKY_POLLING || config.getPersistentQueueMode() == PersistentQueueMode.STICKY_EVENTS) {
+                reaper.start();
+            }
+            isStarted.set(true);
```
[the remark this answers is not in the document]
sbrossie: I am tempted to say 'yes' (with the way our start() method have been written) : The `isInitialized` is needed to ensure only one thread runs the `start()` method; the `isStarted` is needed to ensure nothing can happen prior completed the `start() ` method.

If i do...

```
        if (isStarted.compareAndSet(false, true)) {
            dao.initialize();
            dispatcher.start();
            startQueue();

            if (config.getPersistentQueueMode() == PersistentQueueMode.STICKY_POLLING || config.getPersistentQueueMode() == PersistentQueueMode.STICKY_EVENTS) {
                reaper.start();
            }
        }
```

... then, only one thread will run the block, but the `isStarted` flag is set right away to true prior the code completed: Since this code now potentially pulls entries from the DB (ini sequence), it's important to have other threads wait for the init to have completed.

PR #63, opened 2019-03-13
queue: add documentation

pierre:
Add some implementation notes, since the library is becoming more popular.

Reading the rendered output might be the easiest: https://github.com/killbill/killbill-commons/blob/docs-update/queue/README.md


pierre:
> 1. One thing to add -- unless i messed it -- is that handlers need to be idempotent as they could be retried.

Good point, fixed in 8e3e5501cd82ec2231762c68e3243f4c43be7740.

> 2. Nit: Formatting seems weird, lots of line cut, not sure if this will be reflected after this has been rendered though

It's because I wrote it with new lines at 120 chars or so to avoid overly long lines. Doesn't impact rendering.

PR #64, opened 2019-05-02
Experiment with having 2 threads taking care of queue lifecycle (dispatch + complete)

sbrossie:


sbrossie:
Closed - we decided for a different strategy.

PR #65, opened 2019-05-03
Queue perf optimization

sbrossie:


queue/src/main/java/org/killbill/bus/api/PersistentBusConfig.java
```diff
@@ -87,6 +87,18 @@
     @Description("Max number of dispatch threads to use")
     public abstract int geMaxDispatchThreads();
 
+    @Override
+    @Config("org.killbill.persistent.bus.${instanceName}.lifecycle.dispatch.nbThreads")
+    @Default("1")
+    @Description("Max number of lifecycle dispatch threads to use")
+    public abstract int geNbLifecycleDispatchThreads();
```
pierre: Docker changes: https://github.com/killbill/killbill-cloud/pull/149.

PR #66, opened 2019-05-08
queue: Decouple the reaper threshold with the reaper scheduling period

sbrossie:


queue/src/main/java/org/killbill/queue/DefaultReaper.java
```diff
@@ -27,19 +27,26 @@
 import org.killbill.commons.concurrent.Executors;
 import org.killbill.queue.api.PersistentQueueConfig;
 import org.killbill.queue.api.Reaper;
+import org.skife.config.TimeSpan;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public abstract class DefaultReaper implements Reaper {
 
+    static final long ONE_MINUTES_IN_MSEC = 60000;
```
pierre: Nit: `ONE_MINUTE_IN_MSEC`

PR #68, opened 2019-06-20
jdbi cleanups

pierre:


jdbi/src/main/java/org/killbill/commons/jdbi/guice/DBIProvider.java
```diff
@@ -97,10 +101,28 @@ public void setTimingCollector(final TimingCollector timingCollector) {
         this.timingCollector = timingCollector;
     }
 
+    @Inject(optional = true)
+    public void setStatementRewriter(final StatementRewriter statementRewriter) {
+        this.statementRewriter = statementRewriter;
```
sbrossie: I thought those rewriters were already used by jDBI to rewrite the string template query into real jDBC query. Is there not a collusion or ordering potential issue?

pierre: There's only one in use (`ColonPrefixNamedParamStatementRewriter`). It can be useful to extend that class to update the query from the template before it's processed by ANTLR.

PR #70, opened 2019-07-30
Queue fixes

pierre:
Best is to go commit by commit (see commit messages).


queue/src/test/java/org/killbill/queue/TestReaper.java
```diff
@@ -0,0 +1,254 @@
...
+        // Insert ready entry (not yet reapable)
+        sqlDao.insertEntry(createEntry(1L, CreatorName.get(), null, now, PersistentQueueEntryLifecycleState.AVAILABLE), config.getTableName());
+        // Insert ready entry from another node (not yet reapable)
+        sqlDao.insertEntry(createEntry(2L, "thatOtherNode", null, now, PersistentQueueEntryLifecycleState.AVAILABLE), config.getTableName());
+        // Insert in-processing entry (not yet reapable)
+        sqlDao.insertEntry(createEntry(3L, CreatorName.get(), CreatorName.get(), now, PersistentQueueEntryLifecycleState.IN_PROCESSING), config.getTableName());
+        // Insert in-processing entry from another node (not yet reapable)
+        sqlDao.insertEntry(createEntry(4L, "thatOtherNode", "thatOtherNode", now, PersistentQueueEntryLifecycleState.IN_PROCESSING), config.getTableName());
+        // Insert in-processing entry (reapable)
+        sqlDao.insertEntry(createEntryForReaping(5L, CreatorName.get(), CreatorName.get(), now, config.getReapThreshold().getMillis(), PersistentQueueEntryLifecycleState.IN_PROCESSING), config.getTableName());
```
sbrossie: `(reapable)` -> Not reapable by this node though.

queue/src/test/java/org/killbill/queue/TestReaper.java
```diff
@@ -0,0 +1,254 @@
...
+        // Insert in-processing entry from another node (reapable)
+        sqlDao.insertEntry(createEntryForReaping(6L, "thatOtherNode", "thatOtherNode", now, config.getReapThreshold().getMillis(), PersistentQueueEntryLifecycleState.IN_PROCESSING), config.getTableName());
+
+        // Check ready entries
+        final List<BusEventModelDao> readyEntries = sqlDao.getReadyEntries(now.toDate(), 10, CreatorName.get(), config.getTableName());
+        // One ready entry (STICKY_POLLING mode)
+        assertEquals(readyEntries.size(), 1);
+        assertEquals(readyEntries.get(0).getRecordId(), (Long) 1L);
+
+        // Check ready and in processing entries
```
sbrossie: The last query checks for this node only but this one is global across nodes if i understand correctly.

queue/src/main/java/org/killbill/queue/dao/QueueSqlDao.java
```diff
@@ -75,6 +75,7 @@ long getNbReadyEntries(@Bind("now") Date now,
     List<T> getEntriesLeftBehind(@Bind("max") int max,
                                  @Bind("now") Date now,
                                  @Bind("reapingDate") Date reapingDate,
+                                 @Bind("owner") String owner,
```
sbrossie: I don't think `owner` is used since we fetch all entries and then have the `DBBackedQueue#reapEntries` determine which one are *others* versus *us*.

queue/src/main/java/org/killbill/queue/DBBackedQueue.java
```diff
@@ -271,27 +272,52 @@ public Long execute() throws RuntimeException {
...
+                final Collection<T> entriesToMove = new ArrayList<T>(entriesLeftBehind.size());
+                final List<T> entriesToReInsert = new ArrayList<T>(entriesLeftBehind.size());
+                for (final T entryLeftBehind : entriesLeftBehind) {
+                    // entryIsBeingProcessedByThisNode is a sign of a stuck entry on this node
+                    final boolean entryIsBeingProcessedByThisNode = owner.equals(entryLeftBehind.getProcessingOwner());
+                    // entryCreatedByThisNodeAndNeverProcessed is likely a sign of the queue being late
+                    final boolean entryCreatedByThisNodeAndNeverProcessed = owner.equals(entryLeftBehind.getCreatingOwner()) && entryLeftBehind.getProcessingOwner() == null;
+                    if (entryIsBeingProcessedByThisNode) {
+                        // See https://github.com/killbill/killbill-commons/issues/47
+                        log.warn("{} reapEntries: stuck queue entry {}", DB_QUEUE_LOG_ID, entryLeftBehind);
```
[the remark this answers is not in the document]
pierre: See b2a2693d38a6d97aab8701c72e21b8e86f342e2b.

queue/src/test/java/org/killbill/queue/TestReaper.java
pierre: Good catch, see ead0a8f7b421691999d6489b356ccb9ff43d8db4.

It's actually a left over from d3423bd1af270bfaff52cc69e7ad26c2bef3894c: the entry was legacy-repeable before the change.

queue/src/test/java/org/killbill/queue/TestReaper.java
pierre: Yup

queue/src/main/java/org/killbill/queue/dao/QueueSqlDao.java
pierre: Good catch - it was let over from an initial implementation. See 6279fe9dc0663ef7fcb7d6e3f00cf99bdf4999da.

pierre:
> Overall looks really good.

🤗 

> * Reaper:
>   
>   * Should we revisit our default reaper config ?

Here are the current defaults:

* claim time: 5m
* reap threshold: 10m
* max re-dispatch count: 10
* reaper schedule: 3m

Thinking about the ☁️ auto-scaling use-case, a `STICKY` notification queue will likely never catch-up with these settings (`POLLING` would be more appropriate in a truly dynamic environment). For the bus, assuming it's not late, it might be good enough (we could be a tad more aggressive on the reaper schedule / re-dispatch though).

>   * I had introduced the system property `org.killbill.queue.reap.schedule` to de-corelate the scheduling of the reaper with the period -- should we move it to first class citizen config?

See a5653d70fe6bcce920fc89d2777e75093a1b1bc7.

PR #71, opened 2019-08-01
queue: don't swallow exceptions

pierre:
Signed-off-by: Pierre-Alexandre Meyer <pierre@mouraf.org>


pierre:
> Understand the fix but don't understand the descriptions: None of these exceptions were thrown, so removing them makes sense, but what did that swallow exactly?

See https://github.com/killbill/killbill-commons/pull/71/files#diff-b3279f936caa5ec6c49205041ced7118L55.

PR #74, opened 2019-09-30
Fix for 73

sbrossie:


queue/src/main/java/org/killbill/queue/DefaultQueueLifecycle.java
```diff
@@ -288,4 +300,22 @@ private void sleepSporadically(final long loopTimeMsec) throws InterruptedExcept
...
+
+    private interface RunnableRawCallback {
+        void callback() throws InterruptedException;
+    }
+
+    private void withHandlingRuntimeException(final RunnableRawCallback cb) throws InterruptedException {
+        try {
+            cb.callback();
+        } catch (final DBIException e) {
+            log.warn(String.format("%s: Thread %s got DBIException exception: %s",
```
pierre: image

https://www.slf4j.org/faq.html#logging_performance 😏

sbrossie: Ok, actually, now i remember.. most logs in the `queue` code are written this way:
* Should i only update this new one ?
* Should i  update the whole file?
* Should i  update the whole queue code base?
* Should i  update nothing ?

pierre: Up to you. FWIW my vote would be to update the current file. 👼 

> Leave Things BETTER than you found them. ~ Robert Baden Powell

sbrossie: > FWIW my vote would be to update the current file. 👼

👍 

> Robert Baden Powell 

Really, who is that Dude anyways :-)

sbrossie: @pierre see cb6d28806243f280ee6

PR #78, opened 2020-04-09
queue: Split start sequence into initQueue() and startQueue() for fin…

sbrossie:
…er grain lifecyle

The initQueue() starts everything bus the dispatch lifecycle thread pool and the reaper thread.

This way we can start sending events right after initQueue() but no events are being dispatched until the startQueue() is called -- we expect to plug this step as a last step in our lifecyle after
plugins and KB services have been started.

Also:
* Rework the start/stop sequence and make it more homegeneous between NotificationQueue and PersistentBus
* Fix race conditions during stop sequence

### DIFF
--- a/queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
+++ b/queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
@@ -186,10 +186,10 @@ public class DefaultPersistentBus extends DefaultQueueLifecycle implements Persi
     public void stopQueue() {
         if (isStarted.compareAndSet(true, false)) {
             isInitialized.set(false);
-            super.stopQueue();
             reaper.stop();
-            dao.close();
+            super.stopQueue();
             dispatcher.stop();
+            dao.close();
         }
     }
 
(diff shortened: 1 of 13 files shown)


queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
```diff
@@ -141,35 +141,55 @@ public DefaultPersistentBus(final DataSource dataSource, final Properties proper
...
-
     }
 
     @Override
-    public void stop() {
+    public void stopQueue() {
         if (isStarted.compareAndSet(true, false)) {
             isInitialized.set(false);
+            super.stopQueue();
+            reaper.stop();
```
pierre, line 190: Not sure if it matters but shouldn't the reaper be stopped first (so it doesn't move entries after the queue is stopped)? Also, shouldn't the `dao` be the last one to stop?

queue/src/main/java/org/killbill/notificationq/DefaultNotificationQueue.java
```diff
@@ -395,27 +400,45 @@ public NotificationQueueHandler getHandler() {
         return handler;
     }
 
+    @Override
+    public boolean initQueue() {
+        if (config.isProcessingOff()) {
+            return false;
```
pierre, line 406: Maybe add a log line.

PR #80, opened 2020-04-23
queue: fix InMemoryPersistentBus init sequence

pierre:
This fixes https://github.com/killbill/killbill/issues/1299.

### DIFF
--- a/queue/src/main/java/org/killbill/bus/InMemoryPersistentBus.java
+++ b/queue/src/main/java/org/killbill/bus/InMemoryPersistentBus.java
@@ -109,18 +109,23 @@ public class InMemoryPersistentBus implements PersistentBus {
 
     @Override
     public boolean initQueue() {
-        return false;
-    }
-
-
-    @Override
-    public boolean startQueue() {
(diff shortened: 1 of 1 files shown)


queue/src/main/java/org/killbill/bus/InMemoryPersistentBus.java
```diff
@@ -109,20 +109,25 @@ public void postFromTransaction(final BusEvent event, final Connection connectio
 
     @Override
     public boolean initQueue() {
-        return false;
-    }
-
-
-    @Override
-    public boolean startQueue() {
         if (isInitialized.compareAndSet(false, true)) {
```
sbrossie, line 112: Somehow orthogonal to this change, we could add the check for the `isProcessingOff` -- similarly to `DefaultPersistentBus#initQueue `

pierre: Fixed in e098b7d4b72b6318685826c956fc409ae05ca21d.

Issue #95, opened 2021-08-24
Replace cglib with Byte Buddy for Java 16 compatibility


pierre:
* Fork and re-implement proxies in config-magic
* Re-implement proxies in jdbi

pierre:
For `jdbi`, we can just get rid of `CGLIBDispatchBuilder`, no need to adapt it, as it seems unused. The bulk of the work is to rewrite `SqlObect#buildSqlObject` to use byte-buddy instead of cglib and remove the dependency on `MethodProxy` in the `Handler` interface.

pierre:
See https://github.com/killbill/killbill-commons/pull/99.

PR #96, opened 2021-08-31
Fork config magic

pierre:
config-magic has been unmaintained for years. Fork it and get rid of cglib dependency for Java 16.

See https://github.com/killbill/killbill-commons/issues/95.

Once this is merged, the fork is ready to be used but work is still needed to switch to the new coordinates in the various repos.


sbrossie (approved):
Is this for a KB version 0.22.x ?

PR #97, opened 2021-09-23
queue: revisit shutdown sequence

pierre:
Make timeout configurable to wait for executors threads to complete.

Some handlers can take a long time (e.g. Analytics plugin handlers) and the default hardcoded 5s value isn't always enough in practice. This can result in leaving `IN_PROCESSING` entries on disk if the shutdown sequence doesn't wait for these threads to finish.

As part of this, the shutdown sequence was revisited too. With the previous implementation, even if the shutdown sequence waits for these threads to finish, `IN_PROCESSING` entries could still be left on disk if the completion threads are
stopped before all jobs in the dispatcher queue are done.

I've also bumped the default timeout from 5s to 15s (5s might be a bit aggressive, even for core Kill Bill handlers).

It's a ☕ kind of review 😸 Might be easier to checkout the code...

### DIFF
--- a/queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
+++ b/queue/src/main/java/org/killbill/bus/DefaultPersistentBus.java
@@ -1,8 +1,8 @@
 /*
  * Copyright 2010-2014 Ning, Inc.
  * Copyright 2014-2020 Groupon, Inc
- * Copyright 2020-2020 Equinix, Inc
- * Copyright 2014-2020 The Billing Project, LLC
+ * Copyright 2020-2021 Equinix, Inc
+ * Copyright 2014-2021 The Billing Project, LLC
  *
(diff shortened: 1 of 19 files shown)


sbrossie (approved):
Looks good - but something to keep in mind is that the timeout value `config#getShutdownTimeout()` is not really used a a 'bound' for the shutdown sequence because we use it for each pool (reaper, lifecycle dispatcher, dispatcher, lifecycle completion).

pierre:
> Looks good - but something to keep in mind is that the timeout value `config#getShutdownTimeout()` is not really used a a 'bound' for the shutdown sequence because we use it for each pool (reaper, lifecycle dispatcher, dispatcher, lifecycle completion).

As discussed: while this is true in theory, in practice, the shutdown of reaper, lifecycle dispatcher, and lifecycle completion should virtually be instantaneous.

PR #99, opened 2021-11-02
Replaced cglib with ByteBuddy

vnandwana:
* Removed cglib dependency and added ByteBuddy in killbill-jdbi module.
* Removed MethodProxy from Handler interface and updated all the classes that have implemented this interface.
* Updated SqlObject methods to make use of ByteBuddy instead of cglib for objects creation.
* Deleted CGLIBDispatchBuilder and its unit test class as it is not in use anymore.


pierre:
@vnandwana Could you merge `master` into your branch to pick-up https://github.com/killbill/killbill-commons/commit/4b8d8e647b8028076b5e74cdc27170bda50ca029? That way, CI will run.

Thanks!

pom.xml
```diff
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.kill-bill.billing</groupId>
         <artifactId>killbill-oss-parent</artifactId>
-        <version>0.144.66</version>
+        <version>0.144.72-SNAPSHOT</version>
```
pierre, line 25: Let's separate the work to support jdk16 from the migration to byte-buddy. To replace cglib with byte-buddy, I don't think changes in the oss-parent are required? If so, can you revert this change here so that we can ensure CI tests pass first with byte-buddy, while keeping everything else unchanged?

pierre (changes requested):

pom.xml
vnandwana: It's a good idea to separate the migration to ByteBuddy from JDK 16 support. I did revert this change.

vnandwana: I'm looking into issues that are causing a build failure.

pierre: FYI Looks like it's getting close! Just a few more tests in the jdbi module failing.

pierre (approved):
Very well done!

PR #101, opened 2021-12-08
Excluded duplicate transitive dependency

vnandwana:
* Excluded duplicate ByteBuddy dependency from org.redisson group.
* Updated killbill-oss-parent dependency version.

### DIFF
--- a/clock/pom.xml
+++ b/clock/pom.xml
@@ -51,6 +51,12 @@
             <groupId>org.redisson</groupId>
             <artifactId>redisson</artifactId>
             <scope>test</scope>
+            <exclusions>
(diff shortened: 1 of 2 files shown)


clock/pom.xml
```diff
@@ -51,6 +51,12 @@
             <groupId>org.redisson</groupId>
             <artifactId>redisson</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
```
pierre, line 55: Why do we need to exclude this? In general, we try to avoid exclusions unless absolutely necessary, as they are hard to manage long term.

PR #103, opened 2021-12-10
OutOfMemoryError Fix

vnandwana:
* TypeCache was instantiated on every `buildSqlObject(..)` call. Fixed this by making TypeCache an instance variable.
* Replaced TypeCache with TypeCache.WithInlineExpunction instance, it's a subclass of TypeCache where obsolete references are cleared upon every use of the cache.

### DIFF
--- a/jdbi/src/main/java/org/skife/jdbi/v2/sqlobject/SqlObject.java
+++ b/jdbi/src/main/java/org/skife/jdbi/v2/sqlobject/SqlObject.java
@@ -35,6 +35,7 @@ import com.fasterxml.classmate.TypeResolver;
 import com.fasterxml.classmate.members.ResolvedMethod;
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.TypeCache;
+import net.bytebuddy.TypeCache.WithInlineExpunction;
 import net.bytebuddy.dynamic.loading.ClassLoadingStrategy.Default;
 import net.bytebuddy.implementation.MethodDelegation;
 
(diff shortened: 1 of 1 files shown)


jdbi/src/main/java/org/skife/jdbi/v2/sqlobject/SqlObject.java
```diff
@@ -57,7 +58,7 @@
     {
         final SqlObject so = new SqlObject(buildHandlersFor(sqlObjectType), handle);
 
-        TypeCache<Class<?>> typeCache = new TypeCache<>(TypeCache.Sort.SOFT);
+        TypeCache<Class<?>> typeCache = new WithInlineExpunction<>(TypeCache.Sort.SOFT);
```
pierre, line 61: `typeCache` needs to be static and shared globally, otherwise it's not really a cache :-)

vnandwana: Yes, I agree. This is what I tried first ([please check these changes](https://github.com/killbill/killbill-commons/pull/103/commits/8f9b01707074a0a41764822e17339c4321a76890#diff-9a278e90e2caa8deccb8c68b2da98c3596f7b16cec425af224ecaedde26b4837R49)) and it resulted in some weird behavior ([check this build failure](https://github.com/killbill/killbill-commons/actions/runs/1562949064)).
Currently, with this change, I'm checking if these errors have anything to do with the WithInLineExpunction object type. If it works and for some reason, instance-level instantiation doesn't, then I'm thinking about checking/loading if an instance of TypeCache class type already exists in memory otherwise create a new instance.

PR #104, opened 2022-01-12
queue: Fix issue with reaper in POLLING mode

sbrossie:
Remove the condition which prevents reaped entries to be reinserted (so as to allow them to be processed).


pierre (approved):

kuguobing:
org.killbill.queue.DBBackedQueue:: reapEntries()

} else if (entryCreatedByThisNodeAndNeverProcessed) {
lateEntries.**addAll**(entriesLeftBehind); ==> shoud be : **lateEntries.add(entryLeftBehind)**
} else {

PR #116, opened 2022-04-19
embeddeddb: jdbi: support for mariadb-java-client 3.0.4

pierre:
* `permitMysqlScheme` is now required in the jdbc url to support the `jdbc:mysql` scheme.
* The first byte is at position 1 when calling `blob.getBytes` (per the spec). Not sure why we never noticed the issue, I suspect some codepaths aren't exercised.
* Streaming doesn't rely on the `Integer.MIN_VALUE` hint anymore. I've verified streaming still works as expected (a `StreamingResult` is returned of size 1 and fetches data as needed unlike when `CompleteResult` is used, where everything is loaded in memory).

### DIFF
--- a/embeddeddb/mysql/pom.xml
+++ b/embeddeddb/mysql/pom.xml
@@ -69,6 +69,11 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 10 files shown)


sbrossie (commented):
👍

PR #129, opened 2022-05-14
Lock fix and logs

sbrossie:
DO NOT MERGE

### DIFF
--- a/locker/src/main/java/org/killbill/commons/locker/GlobalLockBase.java
+++ b/locker/src/main/java/org/killbill/commons/locker/GlobalLockBase.java
@@ -59,10 +59,14 @@ public class GlobalLockBase implements GlobalLock {
 
                 if (resetCallback != null && !resetCallback.reset(lockName)) {
                     // We are not the last one using that lock, bail early (AND don't close the connection)...
+                    logger.info("Skip releasing lock (owner re-entered) for " + lockName);
                     return null;
                 }
                 try {
-                    lockDao.releaseLock(connection, lockName);
+                    final boolean status = lockDao.releaseLock(connection, lockName);
+                    if (!status) {
+                        logger.warn("Failed to release global lock for " + lockName);
+                    }
                 } catch (final SQLException e) {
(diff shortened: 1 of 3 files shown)


locker/src/main/java/org/killbill/commons/locker/GlobalLockBase.java
```diff
@@ -59,10 +59,14 @@ public Void execute() throws RuntimeException {
...
                 if (resetCallback != null && !resetCallback.reset(lockName)) {
                     // We are not the last one using that lock, bail early (AND don't close the connection)...
+                    logger.info("Skip releasing lock (owner re-entered) for " + lockName);
                     return null;
                 }
                 try {
-                    lockDao.releaseLock(connection, lockName);
+                    final boolean status = lockDao.releaseLock(connection, lockName);
+                    if (!status) {
+                        logger.warn("Failed to release global lock for " + lockName);
```
pierre, line 68: If we indeed see this, we should update the DAO interface to return the actual value return by MySQL (`0 / null`): https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-lock

locker/src/main/java/org/killbill/commons/locker/GlobalLockerBase.java
```diff
@@ -73,10 +73,14 @@ public GlobalLock execute() throws LockFailedException {
     protected GlobalLock lock(final String lockName) {
         final TryAcquireLockState lockState = lockTable.tryAcquireLockForExistingOwner(lockName);
         if (lockState.getLockState() == ReentrantLock.ReentrantLockState.HELD_OWNER) {
+            logger.info("HELD_OWNER lock for " + lockName);
             return lockState.getOriginalLock();
         }
 
         if (lockState.getLockState() == ReentrantLock.ReentrantLockState.HELD_NOT_OWNER) {
+
+            logger.info("HELD_NOT_OWNER lock for " + lockName);
```
pierre, line 82: Might be worth printing the request id to understand who is holding it?

PR #131, opened 2022-06-03
WIP - Metrics interfaces Proposal

andrenpaes:
### Summary of Proposal
* Removes `Timer` and `Meter` metrics (can be replaced by `Histogram`/`Summary` and `Counter`, respectively)
* Use `double` for metric values,  still compatible enough with `long`, but more flexible in terms of values that can be stored
* Make `Gauge` interface simpler to use, providing `set`, `inc` and `dec` methods, and using a `Supplier` to create the metric.
* Defines the `Summary` metric, which is a metric that's suited for quantile computation on client side (equivalent to Dropwizard's Histogram)
* Changes `Histogram` definition, using metric value range buckets, which is better suited for aggregation
### Labels
* Allows passing labels when creating the metrics, and when providing samples
* For now all labels keys and values are `String` instances.
* Label values would be matched to their keys by the order that their provided (a bit fragile)
* We could evaluate creating a `Label` class for encapsulation (we do that in our golang library)
  * Benefit of making it easier to introduce changes to labels without changing method signatures

### DIFF
--- a/metrics-api/src/main/java/org/killbill/commons/metrics/api/Counter.java
+++ b/metrics-api/src/main/java/org/killbill/commons/metrics/api/Counter.java
@@ -20,21 +20,21 @@ package org.killbill.commons.metrics.api;
 /**
  * An incrementing and decrementing counter metric.
  */
-public interface Counter extends Metric, Counting {
+public interface Counter extends Metric {
 
     /**
      * Increment the counter by {@code n}.
      *
      * @param n the amount by which the counter will be increased
      */
-    void inc(long n);
+    void inc(double n, String ...labels);
 
     /**
-     * Decrement the counter by {@code n}.
-     *
-     * @param n the amount by which the counter will be decreased
+     * Increment the counter by 1.
      */
-    default void dec(final long n) {
-        inc(-n);
+    default void inc(String ...labels) {
+        inc(1, labels);
     }
+
+    double get(String... labels);
 }
--- a/metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
+++ b/metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
@@ -17,7 +17,7 @@
 
 package org.killbill.commons.metrics.api;
 
-import java.util.Map;
+import java.util.function.DoubleSupplier;
 
 public interface MetricRegistry {
 
@@ -28,7 +28,7 @@ public interface MetricRegistry {
      * @param name the name of the metric
      * @return a new or pre-existing {@link Counter}
      */
-    public Counter counter(String name);
+    Counter counter(String name, String ...labelNames);
 
     /**
      * Return the {@link Gauge} registered under this name; or create and register
@@ -38,34 +38,25 @@ public interface MetricRegistry {
      * @param supplier the underlying Gauge
      * @return a new or pre-existing {@link Gauge}
      */
-    <T> Gauge<T> gauge(String name, Gauge<T> supplier);
+    Gauge gauge(String name, String ...labelNames);
+
+    Gauge gauge(String name, DoubleSupplier supplier, String ...labelNames);
 
     /**
      * Return the {@link Histogram} registered under this name; or create and register
-     * a new {@link Histogram} if none is registered.
+     * a new {@link Histogram} if none is registered using the implementation's default
+     * bucket distribution.
      *
      * @param name the name of the metric
      * @return a new or pre-existing {@link Histogram}
      */
-    Histogram histogram(String name);
+    Histogram histogram(String name, String ...labelNames);
 
-    /**
-     * Return the {@link Meter} registered under this name; or create and register
-     * a new {@link Meter} if none is registered.
-     *
-     * @param name the name of the metric
-     * @return a new or pre-existing {@link Meter}
-     */
-    public Meter meter(String name);
+    Histogram histogram(String name, double[] buckets, String ...labelNames);
 
-    /**
-     * Return the {@link Timer} registered under this name; or create and register
-     * a new {@link Timer} if none is registered.
-     *
-     * @param name the name of the metric
-     * @return a new or pre-existing {@link Timer}
-     */
-    Timer timer(String name);
+    Summary summary(String name, String ...labelNames);
+
+    Summary summary(String name, double[] quantiles, String ...labelNames);
 
     /**
      * Removes the metric with the given name.
@@ -73,17 +64,5 @@ public interface MetricRegistry {
      * @param name the name of the metric
      * @return whether or not the metric was removed
      */
-    boolean remove(String name);
-
-    Map<String, ?> getMetrics();
-
-    Map<String, Counter> getCounters();
-
-    Map<String, Histogram> getHistograms();
-
-    Map<String, Gauge<?>> getGauges();
-
-    Map<String, Meter> getMeters();
-
-    Map<String, Timer> getTimers();
+    boolean remove(String name, String ...labelNames);
 }
--- a/metrics-api/src/main/java/org/killbill/commons/metrics/api/Histogram.java
+++ b/metrics-api/src/main/java/org/killbill/commons/metrics/api/Histogram.java
@@ -18,14 +18,14 @@
 package org.killbill.commons.metrics.api;
 
 /**
- * A metric which calculates the distribution of a value.
+ * A metric which calculates the distribution of a value into specific value buckets.
  */
-public interface Histogram extends Metric, Sampling, Counting {
+public interface Histogram extends Metric {
 
     /**
      * Adds a recorded value.
      *
      * @param value the length of the value
      */
-    void update(long value);
+    void update(double value, String ...labels);
 }
--- a/metrics-api/src/main/java/org/killbill/commons/metrics/api/Summary.java
+++ b/metrics-api/src/main/java/org/killbill/commons/metrics/api/Summary.java
@@ -17,21 +17,10 @@
 
 package org.killbill.commons.metrics.api;
 
-public interface Gauge extends Metric {
+/**
+ * A metric which calculates the distribution of a value into quantiles.
+ */
+public interface Summary extends Metric {
 
-    void set(double value, String ...labels);
-
-    void inc(double value, String... labels);
-
-    default void inc(String ...labels) {
-        inc(1, labels);
-    }
-
-    void dec(double value, String... labels);
-
-    default void dec(String ...labels) {
-        dec(1, labels);
-    }
-
-    double get(String ...labels);
+    void update(double value, String ...labels);
 }


metrics-api/src/main/java/org/killbill/commons/metrics/api/Counter.java
```diff
@@ -20,21 +20,21 @@
...
-public interface Counter extends Metric, Counting {
+public interface Counter extends Metric {
 
     /**
      * Increment the counter by {@code n}.
      *
      * @param n the amount by which the counter will be increased
      */
-    void inc(long n);
+    void inc(double n, String ...labels);
```
pierre, line 30: How about `BigDecimal` instead of `double`? That was we always ensure precision?

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
```diff
@@ -38,52 +38,31 @@ public interface MetricRegistry {
...
+    Histogram histogram(String name, double[] buckets, String ...labelNames);
 
-    /**
-     * Return the {@link Timer} registered under this name; or create and register
-     * a new {@link Timer} if none is registered.
-     *
-     * @param name the name of the metric
-     * @return a new or pre-existing {@link Timer}
-     */
-    Timer timer(String name);
```
pierre, line 68: FWIW `Timer` class still around.

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
```diff
@@ -38,52 +38,31 @@ public interface MetricRegistry {
...
+    Histogram histogram(String name, String ...labelNames);
 
-    /**
-     * Return the {@link Meter} registered under this name; or create and register
-     * a new {@link Meter} if none is registered.
-     *
-     * @param name the name of the metric
-     * @return a new or pre-existing {@link Meter}
-     */
-    public Meter meter(String name);
```
pierre, line 59: FWIW `Meter` class still around.

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
```diff
@@ -38,52 +38,31 @@ public interface MetricRegistry {
...
 
     /**
      * Removes the metric with the given name.
      *
      * @param name the name of the metric
      * @return whether or not the metric was removed
      */
-    boolean remove(String name);
-
-    Map<String, ?> getMetrics();
```
pierre, line 78: NACK -- we need to be able to retrieve the metrics (e.g. for `MetricRegistrySerializer` or for `org.killbill.billing.server.metrics.KillbillPluginsMetricRegistry`).

metrics-api/src/main/java/org/killbill/commons/metrics/api/Gauge.java
```diff
@@ -17,12 +17,21 @@
...
 package org.killbill.commons.metrics.api;
 
-public interface Gauge<T> extends Metric {
-
-    /**
-     * Returns the metric's current value.
-     *
-     * @return the metric's current value
-     */
-    T getValue();
```
pierre, line 27: NACK -- or maybe I'm misunderstanding how it would be used.

Unlike other metrics, Gauges are usually created once and callers don't keep a reference to it. Instead, the registry polls the value. Check `org.killbill.billing.util.metrics.JmxAttributeGauge` for instance.

pierre (changes requested):

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
andrenpaes: I was trying to simplify the interface as much as possible. Seems like for `KillbillPluginsMetricRegistry`, `getMetrics` it's just a pass through, I couldn't really find where the methods is being used directly.

I also couldn't find a definition for the serializer. How's that used?

metrics-api/src/main/java/org/killbill/commons/metrics/api/Gauge.java
andrenpaes: The dropwizard model here seems a bit odd, that you have to pass in a Gauge (which is actually a lambda or an anonymous class) to create a Gauge.

My suggestion would be to create something you can manipulate the value (set, inc, dec, etc), or create providing a functional interface (in this case a `DoulbeSupplier`).

metrics-api/src/main/java/org/killbill/commons/metrics/api/Counter.java
andrenpaes: I'm not sure we care enough about precision or rounding for metrics to justify the added interaction complexity `BigDecimal` introduces.

pierre: I think we should still go with `BigDecimal`. This would add more flexibility (e.g. allow decimal values).

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
pierre: `KillbillPluginsMetricRegistry` is a wrapper around `OSGISingleServiceRegistration<MetricRegistry>` (`MetricRegistryServiceRegistration`), which takes care of discovering the services exported by the bundles (currently at most 1, i.e. the `KillBillCodahaleMetricRegistry` one, exported by the `metrics` bundle, although as discussed, we night change that).

This is used in [MetricRegistrySerializer](https://github.com/killbill/killbill-commons/blob/903411dc937c9d91e3f0a9d6bfc7eb201c062fc5/metrics/src/main/java/org/killbill/commons/metrics/servlets/MetricsJacksonModule.java#L245) for instance for the `MetricsServlet` (`/1.0/metrics` endpoint).

metrics-api/src/main/java/org/killbill/commons/metrics/api/Gauge.java
pierre: Agreed, the model is quite odd today. A `Gauge` currently doesn't _have_ a value (the value is retrieved realtime when querying the Gauge).

That being said, **requiring** the caller to _manipulate the value_ (e.g. through _set_) would make it difficult to use in practice. For example, for the `JmxAttributeGauge`, we would need a backend threadpool to `set` the Gauge periodically.

Providing an alternative option (e.g. `DoubleSupplier`) seems necessary.

metrics-api/src/main/java/org/killbill/commons/metrics/api/Counter.java
andrenpaes: No underlying implementation would support BigDecimal

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
andrenpaes: The serializer you pointed is tightly coupled to Dropwizard Metrics. That wouldn't work for something else that doesn't produce all those rates and quantiles metrics.

metrics-api/src/main/java/org/killbill/commons/metrics/api/Gauge.java
andrenpaes: I agree. Would the method for doing so in [MetricRegistry](https://github.com/killbill/killbill-commons/pull/131/files#diff-d5bb2d15c63f242008c51ec78d4641bba62556ecfb15e65f4a9f50e8915a1a1eR43) be enough?

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
pierre: We need to find a way to make these changes _somewhat_ backward compatible though, we wouldn't be able to drop that endpoint entirely.

metrics-api/src/main/java/org/killbill/commons/metrics/api/Gauge.java
pierre: Oh, I missed this! Yup, I think so, to be prototyped 😄

metrics-api/src/main/java/org/killbill/commons/metrics/api/Counter.java
pierre: Talked to @sbrossie -- 👍 for `double`.

metrics-api/src/main/java/org/killbill/commons/metrics/api/MetricRegistry.java
andrenpaes: I was looking into the implementation of the serializer module as a whole, and a lot of what's used wouldn't be available anymore, not just the `getMetrics`, but also the [rates](https://github.com/killbill/killbill-commons/blob/903411dc937c9d91e3f0a9d6bfc7eb201c062fc5/metrics/src/main/java/org/killbill/commons/metrics/servlets/MetricsJacksonModule.java#L176-L181), [mean, min, max, snapshot](https://github.com/killbill/killbill-commons/blob/903411dc937c9d91e3f0a9d6bfc7eb201c062fc5/metrics/src/main/java/org/killbill/commons/metrics/servlets/MetricsJacksonModule.java#L138-L141) and company, which are all Dropwizard specifc.

A simple option would be to add `getSerializer` method to MetricRegistry. That Jackson module then can be easily changed to use that serializer. But that would introduce a dependency to jackson in the metrics piece, most likely.

Another option is having the implementation bundles exposing a Servlet that would be mounted?

Thoughts?

pierre: Ignoring this servlet for a moment, we still need a way to _get_ the metrics out of the registry though, for the _sinks_ (_reporters_)?

e.g. 
https://github.com/killbill/killbill-platform/blob/786cca00de17da3af74c0d8872268644c8d353a0/osgi-bundles/bundles/prometheus/src/main/java/org/killbill/billing/osgi/bundles/prometheus/KillBillCollector.java#L55-L69

Otherwise, the bundle implementing the registry must take care of the reporting as well, but this seems like an unnecessary restriction. Also, I'm not sure how we would implement metrics-based healthchecks, like https://github.com/killbill/killbill-commons/blob/903411dc937c9d91e3f0a9d6bfc7eb201c062fc5/jdbi/src/main/java/org/killbill/commons/jdbi/hikari/KillBillHealthChecker.java#L106



Also, looking at other similar projects:
 https://github.com/Netflix/servo/blob/98d4bccbbd4699128c8ee60e67aab1a14b0c410e/servo-core/src/main/java/com/netflix/servo/MonitorRegistry.java#L30

andrenpaes: > Otherwise, the bundle implementing the registry must take care of the reporting as well

I think that should be the case. Otherwise, we'll have to expose all possible details about a metric in the kb metrics api interface. The problem being that those details are different, depending on the implementation used for metrics.

As an example, Dropwizard Meters (which are equivalent to regular counters in other libraries), keep a lot more info in memory than other implementations (like rates). If we have separate sinks for a dropwizard Meter, than sending that info to a storage might make sense, but for prometheus, not so.

I think the issue in general, is that the current interfaces, and allowing separate sink implementations, make it very hard to have an implementation that's not based in Dropwizard, like prometheus.

We should talk more about this in person to see where we go from here.

PR #134, opened 2022-07-07
1615 killbill utils cache

xsalefter:
Add cache implementation.

Note for `TestInMemoryEventBus` :

- Test methods mixed with "slow" and "fast" test group. Decided to make it "fast" because it supposed to be "fast" tests
- add `alwaysRun = true` because sometimes test failed with NPE (for example NPE in line 153, or 158)

### DIFF
--- a/queue/src/test/java/org/killbill/bus/TestInMemoryEventBus.java
+++ b/queue/src/test/java/org/killbill/bus/TestInMemoryEventBus.java
@@ -35,7 +35,7 @@ public class TestInMemoryEventBus {
     private TestEventBusBase testEventBusBase;
     private PersistentBus busService;
 
-    @BeforeClass(groups = "fast", alwaysRun = true)
+    @BeforeClass(groups = "fast")
     public void beforeClass() throws Exception {
         busService = new InMemoryPersistentBus(new PersistentBusConfig() {
             @Override
--- a/queue/src/test/java/org/killbill/bus/TestInMemoryEventBus.java
+++ b/queue/src/test/java/org/killbill/bus/TestInMemoryEventBus.java
@@ -35,7 +35,7 @@ public class TestInMemoryEventBus {
     private TestEventBusBase testEventBusBase;
     private PersistentBus busService;
 
-    @BeforeClass(groups = "fast")
+    @BeforeClass(groups = "fast", alwaysRun = true)
     public void beforeClass() throws Exception {
         busService = new InMemoryPersistentBus(new PersistentBusConfig() {
             @Override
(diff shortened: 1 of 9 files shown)


queue/src/test/java/org/killbill/bus/TestInMemoryEventBus.java
```diff
@@ -35,7 +35,7 @@ public class TestInMemoryEventBus {
     private TestEventBusBase testEventBusBase;
     private PersistentBus busService;
 
-    @BeforeClass(groups = "fast")
+    @BeforeClass(groups = "fast", alwaysRun = true)
```
pierre, line 38: 🤔 Saw your comment regarding the NPE. ack for now, but I suspect something else is happening...

utils/src/main/java/org/killbill/commons/utils/cache/DefaultCache.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public V getOrLoad(final K key, final Function<K, V> loader) {
+        final V value = get(key);
+        return value == null ? loader.apply(key) : value;
+    }
+
+    @Override
+    public void put(final K key, final V value) {
+        evictExpireEntry(key);
```
pierre, line 133: You probably don't need this here since you are replacing the key anyways?

utils/src/main/java/org/killbill/commons/utils/cache/TimedValue.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final TimedValue<?> that = (TimedValue<?>) o;
+        return expireTime == that.expireTime && value.equals(that.value);
```
pierre, line 60: I think your code technically allows `null` value today, so you might need an NPE check here (or a precondition above).

pierre (approved):

queue/src/test/java/org/killbill/bus/TestInMemoryEventBus.java
xsalefter: Hmm.. maybe because it was mixed with "slow" and "fast" group? And now since every test in `TestInMemoryEventBus` is "fast", it should be fine without `alwaysRun = true` ? 

Let me run again several times locally and report here what happened.

utils/src/main/java/org/killbill/commons/utils/cache/DefaultCache.java
xsalefter: Yes, you're right.

utils/src/main/java/org/killbill/commons/utils/cache/TimedValue.java
xsalefter: No, `null` should not allowed. I update the code and added tests as well.

pierre (approved):

PR #138, opened 2022-07-21
1615 remove guava part 2

xsalefter:
- `DataSourceProvider` : Still use Guice because of there's optional injection.

- `ListAnnotationResolverTest` : make it public with public test method

- `LowerToCamelBeanMapper` : `CaseFormat` functionality replaced by `#toCamelCase()` and `#toSnakeCase()` in `util`'s `Strings` . Forking `CaseFormat` take a lot of code from Guava.

question: Is there any way I can run `TestContainerFactory` test?


jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
```diff
@@ -24,20 +24,22 @@
...
 import org.killbill.commons.jdbi.hikari.KillBillMetricsTrackerFactory;
 import org.killbill.commons.metrics.api.MetricRegistry;
+import org.killbill.commons.utils.annotation.VisibleForTesting;
 import org.skife.config.TimeSpan;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.annotations.VisibleForTesting;
+// FIXME-1615: Still using guice because metricRegistry and healthCheckRegistry need optional injection
+//   not sure what going to do with this. Probably leave this as is.
```
pierre, line 41: Maybe something like this would work?

```
@Inject
@Nullable
public void setMetricsRegistry(final MetricRegistry metricRegistry)
```

or

```
@Inject
public void setMetricsRegistry(final Optional<MetricRegistry> metricRegistry)
```

If not, add the arguments to the `Inject` constructor. Users of `DataSourceProvider` could `bind` a `null` provider for these arguments.

jdbi/src/test/java/org/skife/jdbi/v2/TestContainerFactory.java
```diff
@@ -32,16 +32,16 @@
...
 import java.util.Set;
 import java.util.UUID;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
 
+// FIXME-1615 : How to run this test?
```
pierre, line 44: I think only the `@JDBITests` are run when invoking `mvn test -Pjdbi` (check the pom.xml). I think these `@JDBIQuarantineTests` tests are disabled and we could remove them now.

pierre (approved):

jdbi/src/main/java/org/killbill/commons/jdbi/guice/DataSourceProvider.java
xsalefter: @pierre fixed. FYI the one that works is something like this (and this is what we have now):
```java
    @Inject
    public void setMetricsRegistry(@Nullable final MetricRegistry metricRegistry) {
        this.metricRegistry = metricRegistry;
    }
```

pierre:
👍

Issue #4, opened 2014-06-25
TestDOTBuilder is failing


kevinpostlewaite:
From Pierre:
" it's a bug indeed in the test. The ordering of key/value pairs is not deterministic, i.e. node_16 [color=grey style=filled label=INIT]; becomes node_16 [style=filled color=grey label=INIT]; on your machine. Funny we never ran into it, maybe you have a different Java version?"

Java version: Java(TM) SE Runtime Environment (build 1.8.0_05-b13)

GM16799:surefire-reports kpostlewaite$ cat testng-results.xml 
<?xml version="1.0" encoding="UTF-8"?>
<testng-results skipped="0" failed="1" total="4" passed="3">
  <reporter-output>
  </reporter-output>
  <suite name="Command line suite" duration-ms="295" started-at="2014-06-25T00:53:18Z" finished-at="2014-06-25T00:53:18Z">
    <groups>
      <group name="fast">
        <method signature="TestDOTBuilder.testGenerator()[pri:0, instance:org.killbill.automaton.dot.TestDOTBuilder@6615435c]" name="testGenerator" class="org.killbill.automaton.dot.TestDOTBuilder"/>
        <method signature="TestStateMachine.testLinkedStateMachines()[pri:0, instance:org.killbill.automaton.TestStateMachine@5315b42e]" name="testLinkedStateMachines" class="org.killbill.automaton.TestStateMachine"/>
        <method signature="TestStateMachine.testStateMachine()[pri:0, instance:org.killbill.automaton.TestStateMachine@5315b42e]" name="testStateMachine" class="org.killbill.automaton.TestStateMachine"/>
        <method signature="TestStateMachine.testStateTransition()[pri:0, instance:org.killbill.automaton.TestStateMachine@5315b42e]" name="testStateTransition" class="org.killbill.automaton.TestStateMachine"/>
      </group> 
    </groups>
    <test name="Command line test" duration-ms="295" started-at="2014-06-25T00:53:18Z" finished-at="2014-06-25T00:53:18Z">
      <class name="org.killbill.automaton.TestStateMachine">
        <test-method status="PASS" signature="testLinkedStateMachines()[pri:0, instance:org.killbill.automaton.TestStateMachine@5315b42e]" name="testLinkedStateMachines" duration-ms="201" started-at="2014-06-24T17:53:18Z" finished-at="2014-06-24T17:53:18Z">
          <reporter-output>
          </reporter-output>
        </test-method> 
[...]

PR #142, opened 2022-08-08
Fix TestDOTBuilder is failing

xsalefter:
simplify `@Test` annotation and add diamond to `new TreeSet<>()` declaration.

Commit that actually fix the problem in #4 is previous commit: 3979ffe2.


sbrossie (approved):
Should we then close https://github.com/killbill/killbill-commons/issues/4 ?

xsalefter:
Yes, we can close it.

pierre (approved):

Issue #113, opened 2022-03-03
Address Spotbugs errors


pierre:
1. Enforce the Spotbugs checks by enabling the property `check.fail-spotbugs`
2. Remove the various `check.spotbugs-exclude-filter-file` properties and associated `spotbugs-exclude.xml` files
3. Remove the `@SuppressFBWarnings` annotations
4. Address the errors.

xsalefter:
PR will be per-module basis, and this is the order:

```
[INFO] killbill-commons
[INFO] Kill Bill Clock classes and interfaces for killbill and killbill-commons
[INFO] Kill Bill library of java.util.concurrent wrappers
[INFO] Kill Bill config-magic
[INFO] Kill Bill library of embedded dbs
[INFO] Kill Bill library of embedded dbs: common
[INFO] Kill Bill library of embedded dbs: H2
[INFO] Kill Bill library that provides utility methods for core Java(tm) classes
[INFO] Kill Bill library of embedded dbs: MySQL
[INFO] Kill Bill library of embedded dbs: PostgreSQL
[INFO] killbill-metrics-api
[INFO] Kill Bill utilities for jdbi
[INFO] Kill Bill distributed locking library
[INFO] Kill Bill Persistent Queues (bus/notificationQ)
[INFO] killbill-metrics
[INFO] Kill Bill utilities to build webservices
[INFO] Kill Bill xml utils for loading jaxb xml object tree
[INFO] Kill Bill Automaton (state machines automaton)
```

xsalefter:
As PRs for `xmlloader` and `automaton` merged, all spotbugs issue in this repository already fixed.

PR #148, opened 2022-09-20
remove spotbugs warning in jdbi module

xsalefter:
Related issue: https://github.com/killbill/killbill-commons/issues/113

- I removed this line and [comment](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:oss-113-spotbugs-jdbi?expand=1#diff-d82ba7b2bcee63ab97248c98f5337db46f74aef333898db70e415157786abac3L119): 
  ```java
  // Ensure we don't store an identity map entry which has a hard reference
  // to the key (through the value) by copying the value, avoids potential memory leak.
  found.put(cache_key, name == cache_key ? new String(name) : name);
  ```
  I'm not sure that `new String(name)` is necessary, but then again, I never experiencing any ".... identity map entry which has a hard reference that causing memory leak", so any thought are welcome.

- `LowerToCamelBeanMapper` get refactored, and verify against `TestLowerToCamelBeanMapper`.

- `BeanMapper` get refactored, add new test `TestBeanMapper`.

### DIFF
--- a/jdbi/src/main/java/org/killbill/commons/jdbi/mapper/LowerToCamelBeanMapper.java
+++ b/jdbi/src/main/java/org/killbill/commons/jdbi/mapper/LowerToCamelBeanMapper.java
@@ -44,14 +44,12 @@ import org.killbill.commons.utils.Strings;
 import org.skife.jdbi.v2.StatementContext;
 import org.skife.jdbi.v2.tweak.ResultSetMapper;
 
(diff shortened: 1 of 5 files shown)


xsalefter:
Closed. Changes in this class made a lot of failed test in `-Pjdbi` test that it is better to [replace this with new one](https://github.com/killbill/killbill-commons/pull/149).

PR #149, opened 2022-09-23
Commons 113 jdbi

xsalefter:
Replacement of [this PR](https://github.com/killbill/killbill-commons/pull/148) : See close comment message on this PR of why it closed.

This PR not "resolve all `SuppressFBWarnings` classes", nor "delete `spotbugs-exclude.xml` file". Instead, this PR add more lines in `spotbugs-exclude.xml`. Probably the only reason why you need accept this PR is because now we have more contextual reasons about why some classes/types/patterns are excluded.

1. Use `spotbugs-exclude.xml` instead of `SuppressFBWarnings` annotation, except we can't add exclusion in `spotbugs-exclude.xml` (e.g: because bug in Spotbug itself)

2. Add justification as comment in XML file to explain why it needed: [DateTime](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R57), [PreparedBatchPart](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R57), [NoOpLog](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R127)

3. For every classes/types/methods/patterns that are nearly impossible or took a lot of resources to fix, I added them to `spotbugs-exclude.xml`. For example: [java.lang.Class](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R28), String template [STGroup](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R67), [MetrixRegistry](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R72).

4. For every classes/types/methods/patterns that I'm not sure/not know how exactly to fix, I added with `FIXME` mark. In this repository, it is `FIXME-COMMONS-113`. For example: [OracleReturning](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R100), [Log4JLog](https://github.com/killbill/killbill-commons/compare/work-for-release-0.23.x...xsalefter:commons-113-jdbi?expand=1#diff-05af9331657d4cadf48bedb050cd04b80dbddd81619a8ce6ea4e9dbd32b2ffd1R141). Any suggestion for how to fix them are very welcome.

5. I removed [this line](https://github.com/killbill/killbill-commons/pull/149/files#diff-d82ba7b2bcee63ab97248c98f5337db46f74aef333898db70e415157786abac3R112):
  ```java
    // Ensure we don't store an identity map entry which has a hard reference
    // to the key (through the value) by copying the value, avoids potential memory leak.
    found.put(cache_key, name == cache_key ? new String(name) : name);
  ```
  
  I'm not sure that `new String(name)` is necessary, but then again, I never experiencing any ".... identity map entry which has a hard reference that causing memory leak", so any thought are welcome.


jdbi/spotbugs-exclude.xml
```diff
@@ -15,17 +15,133 @@
...
+        <!--
+        Try simple fix by wrapping with new Arraylist<>(cleanables) result with a lot of:
+        ".... org.skife.jdbi.v2.TestIterator.<xxx> .... java.lang.AssertionError: Handle was not closed correctly!"
+        when running test with: "mvn -Pjdbi test verify".
+        -->
+        <Method name="getCleanables" />
+        <Bug pattern="EI_EXPOSE_REP" />
+    </Match>
+
+    <!-- FIXME-COMMONS-113 - Is it worth investigating? as:
```
pierre, line 100: Probably not 😄 

It was deprecated a long time ago (https://github.com/jdbi/jdbi/commit/1390cfae67dec84f542241f89a07830697bd5b52) and is even still beta on jdbi3 (https://github.com/jdbi/jdbi3-oracle12/blob/0a9ed5c63787f3a8da7ce46fe9323c85d839d7e4/src/main/java/org/jdbi/v3/oracle12/OracleReturning.java#L50).

Regardless, there is much more work needed for us to support Oracle... You can delete the class for now.

jdbi/spotbugs-exclude.xml
```diff
@@ -15,17 +15,133 @@
...
+    </Match>
+
+    <Match>
+        <!-- justification: Iterator created inside for-loop, NoSuchElementException unlikely to happen -->
+        <Class name="org.skife.jdbi.v2.sqlobject.BatchHandler$5" />
+        <Method name="next" />
+        <Bug pattern="IT_NO_SUCH_ELEMENT" />
+    </Match>
+
+    <!-- FIXME-COMMONS-113 - Need investigate, but probably nothing we can do about this -->
```
pierre, line 141: You can delete as well and also remove the dependency in the pom.xml (https://github.com/killbill/killbill-commons/blob/77463d9f0505a2ada5aeed5f93a910a703015a86/jdbi/pom.xml#L102-L106).

Kill Bill uses slf4j exclusively as the logging facade.

jdbi/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -108,22 +101,20 @@ public String locate(String name, StatementContext ctx) throws Exception
...
                 if (in_stream == null) {
                     in_stream = ctx.getSqlObjectType().getResourceAsStream(filename);
                 }
             }
 
             if (in_stream == null) {
-                // Ensure we don't store an identity map entry which has a hard reference
-                // to the key (through the value) by copying the value, avoids potential memory leak.
-                found.put(cache_key, name == cache_key ? new String(name) : name);
+                found.put(cacheKey, name);
```
pierre, line 112: Are you sure about this? See https://github.com/jdbi/jdbi/issues/93 and https://github.com/jdbi/jdbi/pull/125.

pierre (approved):
Just a few comments.

> 1. Use `spotbugs-exclude.xml` instead of `SuppressFBWarnings` annotation, except we can't add exclusion in `spotbugs-exclude.xml` (e.g: because bug in Spotbug itself)
For reference, which bug are you referring to?

PR #157, opened 2022-11-09
Commons 113 spotbugs in xmlloader and automaton

xsalefter:
Some notes: 
- I don't think we need copy constructor for [jaxb classes](https://github.com/killbill/killbill-commons/commit/b23e678d9ccbce7a86ddd742a63d7e88304ad13e).
- I call `clone()` to getter/setter to array [based attributes](https://github.com/killbill/killbill-commons/commit/6551d89dbc8b0120e9668ba103fc3db337e11f2e).
- This is last PR for #113 in killbill-commons.

### DIFF
--- a/automaton/src/main/java/org/killbill/automaton/DefaultStateMachineConfig.java
+++ b/automaton/src/main/java/org/killbill/automaton/DefaultStateMachineConfig.java
@@ -35,9 +35,6 @@ import javax.xml.bind.annotation.XmlRootElement;
 
 import org.killbill.xmlloader.ValidationErrors;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
-@SuppressFBWarnings({"EI_EXPOSE_REP", "EI_EXPOSE_REP2"})
 @XmlRootElement(name = "stateMachineConfig")
 @XmlAccessorType(XmlAccessType.NONE)
 public class DefaultStateMachineConfig extends StateMachineValidatingConfig<DefaultStateMachineConfig> implements StateMachineConfig, Externalizable {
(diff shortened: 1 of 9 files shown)


automaton/src/main/java/org/killbill/automaton/DefaultStateMachineConfig.java
```diff
@@ -99,15 +96,15 @@ public StateMachine getStateMachineForState(final String stateName) throws Missi
     }
 
     public DefaultStateMachine[] getStateMachines() {
-        return stateMachines;
+        return stateMachines == null ? null : stateMachines.clone();
```
pierre, line 99: This makes a shallow copy though - is this what we want?

xsalefter: @pierre yeah you're correct, (especially in automaton related classes) I put that mainly to eliminate spotbugs errors. I'm thinking about something else (copy constructor, etc) but I'm not sure how far we should go:

- Should copy constructor sufficient?
- Since most of implementation classes contains `Externalizable`, should we [concern about security](https://christian-schneider.net/JavaDeserializationSecurityFAQ.html) issue as well?
- If we do concern about security, which approach should we use? I read about [serial killer](https://github.com/ikkisoft/SerialKiller) for blacklist/whitelisting `ObjectInputStream`. But should we go with another options? (and if we should, what options do you have in mind?)
- Or should we remove `clone()` because it resource waste without any advantages anyway? And we should also ignore security issue because it handle somewhere else? (I check it used in `killbill/payment`, but I'm not checking thoroughly each classes to make sure that we handle security issue)

WDYT?

pierre: It's mostly used in payment module (internal use-case). I'm thinking we just ignore the spotbug warning, thats could be a lot of work for not many obvious benefits...

pierre (approved):

PR #167, opened 2023-06-09
send repository_dispatch event

xsalefter:
Original issue: https://github.com/killbill/killbill-oss-parent/issues/751

~See complete explanation here: https://github.com/killbill/gh-actions-shared/pull/41~

The `GITHUB_TOKEN` secret variables need following access:
- metadata:read
- actions:read&write
- contents:read&write

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,11 @@ on:
         description: 'tag to (re-)perform (in case of release:perform failure)'
         required: false
         default: ''
+      release_oss_parent:
+        description: 'Release killbill-oss-parent automatically?'
+        required: true
+        default: true
+        type: boolean
 
 env:
   MAVEN_FLAGS: "-B --no-transfer-progress"
@@ -93,3 +98,11 @@ jobs:
           echo "scm.url=scm\:git\:git@github.com\:${GITHUB_REPOSITORY}.git" > release.properties
           echo "scm.tag=${{ github.event.inputs.perform_version }}" >> release.properties
           mvn ${MAVEN_FLAGS} release:perform
+      - name: Send Repository Dispatch Event
+        if: github.event.inputs.release_oss_parent == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GH_WORKFLOW_PAT }}
+        run: |
+          PROJECT_VERSION=$(git describe --abbrev=0 | cut -d '-' -f 3)
+          gh workflow -R killbill/killbill-oss-parent run release.yml -f commons_version=${PROJECT_VERSION}
+


.github/workflows/release.yml
```diff
@@ -93,3 +99,22 @@ jobs:
           echo "scm.url=scm\:git\:git@github.com\:${GITHUB_REPOSITORY}.git" > release.properties
           echo "scm.tag=${{ github.event.inputs.perform_version }}" >> release.properties
           mvn ${MAVEN_FLAGS} release:perform
+      - name: Send Repository Dispatch Event
+        # We not send any repository dispatch if parent_version and perform_version set
+        if: success() && github.event.inputs.send_repository_dispatch == 'yes' && github.event.inputs.parent_version == '' && github.event.inputs.perform_version == ''
```
pierre, line 104: I don't think the `github.event.inputs.parent_version == ''` check is necessary. Even if the oss-parent version was updated here, we want to trigger the release of oss-parent (with the new version of commons).

I am also wondering if we shouldn't just check for `success() && github.event.inputs.send_repository_dispatch == 'yes'`. If `release:prepare` is successful (new tag pushed) but `release:perform` fails (e.g.m sonatype timeout), this is when we re-run and specify `perform_version`. In that case, we will want the oss-parent to be released with the new version of commons.

.github/workflows/release.yml
```diff
@@ -93,3 +99,22 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_SECRET }}
+          OWNER: xsalefter
```
pierre, line 107: `killbill` 😏

.github/workflows/release.yml
```diff
@@ -93,3 +99,22 @@ jobs:
...
+          REPO: kb-oss-parent
```
pierre, line 108: `killbill-oss-parent`

.github/workflows/release.yml
```diff
@@ -93,3 +99,22 @@ jobs:
...
+          # using double quote in request body because we need to parse variable from string
+        run: |
+          VERSION_WITH_SNAPSHOT=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
```
pierre, line 111: I think this is the wrong version. If version is `1.0.0-SNAPSHOT`, after `release:prepare` tags `1.0.0`, it sets the version to `1.0.1-SNAPSHOT`. I suspect you would see `1.0.1` here. Try `git describe --abbrev=0 | cut -d '-' -f 3`

.github/workflows/release.yml
xsalefter: Yeah you're right. We have `github.event.inputs.send_repository_dispatch == 'yes'`, and let this alone to be a control whether we want to send repository_dispatch or not.

.github/workflows/release.yml
xsalefter: copy-paste things.. :eyes:

pierre (approved):

pierre (approved):

PR #174, opened 2024-10-24
Strengthen nodes info changes 

vnandwana:
Added a property to generate a UUID as the creator name when QUEUE_CREATOR_NAME is not set.


pierre (approved):

queue/src/main/java/org/killbill/CreatorName.java
```diff
@@ -30,6 +30,9 @@ public class CreatorName {
 
     private static final Object lock = new Object();
 
+    // If set to true and QUEUE_CREATOR_NAME is not set, generate a UUID instead of using default hostname.
+    static final String AUTO_GENERATED_QUEUE_CREATOR_NAME = "org.killbill.queue.creator.autoGeneratedName";
```
pierre, line 34: Let's make sure to update the docs (/cc @reshmabidikar)

PR #177, opened 2025-02-03
locker: With potential connection leak when acquiring lock

sbrossie:


pierre (approved):

sbrossie:
I am closing this PR as this completely wrong: The DB `connection` cannot be released when the lock is held because closing the DB connection will release the lock - this is a property we exploit to guarantee that if a node goes down while holding a lock, it does not remain locked forever.

The closing of the connection will happen when the caller calls `GlobalLock#release`, typically from within a finally statement within the code.

PR #179, opened 2025-04-14
Update default queue mode to POLLING

reshmabidikar:
Updated default queue mode to `POLLING` as discussed [here](https://github.com/killbill/killbill-aviate-plugin/issues/40#issuecomment-2798801111). 

~~Edit: Moving to `DRAFT` as there are [ci failures](https://github.com/killbill/killbill-commons/actions/runs/14442833600).~~

Moving back to `IN_REVIEW` as there are no ci failures now.

### DIFF
--- a/config-magic/pom.xml
+++ b/config-magic/pom.xml
@@ -66,7 +66,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <!-- Don't run JUnit tests -->
-                    <groups />
+                    <groups/>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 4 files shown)


sbrossie (commented):
If we change bus configuration by default, should we do the same for notificationQ: https://github.com/killbill/killbill-commons/blob/master/queue/src/main/java/org/killbill/notificationq/api/NotificationQueueConfig.java#L66 ?

Also, let's keep track of it so we have it in our release notes for next Kill Bill 's release.

reshmabidikar:
> If we change bus configuration by default, should we do the same for notificationQ: https://github.com/killbill/killbill-commons/blob/master/queue/src/main/java/org/killbill/notificationq/api/NotificationQueueConfig.java#L66 ?


Changed the default config for notification queues to `POLLING` in 89da219

pierre (approved):

